PR #41, opened 2022-06-06
Chinese translation

DevCoz:
Chinese translation


Skittyblock:
Thank you for adding some translations! I'll try and make this an easier process in the future, so my apologies that I don't have anything set up yet.

DevCoz:
Yes, thank you

PR #46, opened 2022-06-09
Updated Chinese translation

DevCoz:


Shared/zh-Hans.lproj/Localizable.strings
```diff
@@ -138,6 +138,15 @@
 
 "SAVE_TO_PHOTOS" = "保存到照片";
 
+// History
+"HISTORY" = "历史";
+"%@_AGO" = "%@ ago"; // x 天前
```
beerpiss, line 143: You're translating the comment here, not the actual string

beerpiss, line 143: You're translating the comment and not the actual string here.

beerpiss (changes requested):

Shared/zh-Hans.lproj/Localizable.strings
```diff
@@ -25,9 +25,9 @@
 "DOWNLOADED" = "已下载";
 "DOWNLOAD_QUEUE" = "下载队列";
 "CANCEL_DOWNLOAD" = "取消下载";
-"REMOVE_DOWNLOAD" = "移除下载";
-"REMOVE_ALL_DOWNLOADS" = "移除所有下载";
-"REMOVE_DOWNLOADS" = "移除下载";
+"REMOVE_DOWNLOAD" = "删除下载";
+"REMOVE_ALL_DOWNLOADS" = "删除所有下载";
+"REMOVE_DOWNLOADS" = "删除下载";
 "REMOVE_DOWNLOADS_CONFIRM" = "你确定要删除所有选定的下载吗？?";
```
beerpiss, line 31:
```suggestion
"REMOVE_DOWNLOADS_CONFIRM" = "你确定要删除所有选定的下载吗？";
```
There are two question marks here.

Shared/zh-Hans.lproj/Localizable.strings
```diff
@@ -61,14 +61,14 @@
 // Browse
 "BROWSE" = "浏览";
 "BROWSE_NO_SOURCES" = "没有源";
-"BROWSE_NO_SOURCES_TEXT" = "配置您的源列表或在Aidoku中打开源以开始使用";
+"BROWSE_NO_SOURCES_TEXT" = "自定义源列表或\使用 Aidoku 打开源以开始使用";
```
beerpiss, line 64: There is a backslash in the string. Were you trying to add a line break?

Shared/zh-Hans.lproj/Localizable.strings
```diff
@@ -138,6 +138,15 @@
 
 "SAVE_TO_PHOTOS" = "保存到照片";
 
+// History
+"HISTORY" = "历史";
+"%@_AGO" = "%@ ago"; // x 天前
+"REMOVE_CHAPTER_HISTORY_TEXT" = "这将删除此章节历史记录。是否确实要继续?";
```
beerpiss, line 144:
```suggestion
"REMOVE_CHAPTER_HISTORY_TEXT" = "这将删除此章节历史记录。是否确实要继续？";
```
You should keep the question mark variant consistent, make it either `U+FE56 ？SMALL QUESTION MARK` or `U+003F ? QUESTION MARK`

beerpiss (changes requested):

Shared/zh-Hans.lproj/Localizable.strings
beerpiss: If you decide to go with the former, you should also fix `RESTORE_BACKUP`'s question mark.

Shared/zh-Hans.lproj/Localizable.strings
DevCoz: OK, thanks for pointing that out.

Shared/zh-Hans.lproj/Localizable.strings
```diff
@@ -138,6 +138,15 @@
 
 "SAVE_TO_PHOTOS" = "保存到照片";
 
+// History
+"HISTORY" = "历史";
+"%@_AGO" = "%@ ago"; // x days ago
```
beerpiss, line 143:
```suggestion
"%@_AGO" = "%@前"; // x days ago
```
Is keeping the string untranslated intentional? In usage, `%@` will be replaced with a proper phrase for your language so you don't need to worry about things not making sense.

beerpiss (changes requested):

PR #172, opened 2023-02-10
Fix delete locked category

JBLOGNON:
Fixing an issue when deleting a locked category.

Now when a category is deleted, it is also removed from the list of locked categories (if it was in).

Therefore, if you create a category with the same name as a category that was locked (and has been deleted), it will not be directly locked when created.


Skittyblock:
I'm just gonna commit changes to this pr to fix it instead of reviewing it so it's quicker, but the main issue is that you got the core data contexts messed up and it's running a blocking function and then saving two contexts (one of which doesn't matter) since you removed the `async` on `removeCategory` instead of turning it into a function that takes a context. Technically it probably works, but it's not correct.

PR #177, opened 2023-02-19
feat: make description collapsible

KaBankz:
First time using swift in prod, so hope this is good enough.

I tested the changes in a simulator and all was working well.


Skittyblock:
closing this since i unfortunately decided i didn't want to have it, reasons being:
- default apple apps (like app store) typically don't have this, and that's what it was modeled after
- the less text could possibly cover up text if the final line stretches the full width

PR #185, opened 2023-02-26
Update the library view to include downloading

### LINKED ISSUE #182: Batch download [Unread/All] from Library
Would love to request this feature if possible. Designed this UI for easy understanding.

Download
Download_pop

zabroseric:
Resolves #182

### DIFF
--- a/iOS/Scenes/Library/LibraryViewController.swift
+++ b/iOS/Scenes/Library/LibraryViewController.swift
@@ -1012,10 +1012,10 @@ extension LibraryViewController {
             }
 
 
-            let downloadAllAction = UIAction(title: "All") { action in
+            let downloadAllAction = UIAction(title: NSLocalizedString("ALL", comment: "")) { action in
                 self.downloadAllAction(manga: manga.toManga())
             }
-            let downloadUnreadAction = UIAction(title: "Unread") { action in
+            let downloadUnreadAction = UIAction(title: NSLocalizedString("UNREAD", comment: "")) { action in
                 self.downloadUnreadAction(manga: manga.toManga())
             }
 
--- a/iOS/Scenes/Library/LibraryViewController.swift
+++ b/iOS/Scenes/Library/LibraryViewController.swift
@@ -944,19 +944,6 @@ extension LibraryViewController {
         super.collectionView(collectionView, didUnhighlightItemAt: indexPath)
     }
 
-    func downloadAll(manga: Manga) async {
-        let chapters = await CoreDataManager.shared.getChapters(sourceId: manga.sourceId, mangaId: manga.id)
-        DownloadManager.shared.download(chapters: chapters, manga: manga)
-    }
-
-    func downloadUnread(manga: Manga) async {
-        let readingHistory = await CoreDataManager.shared.getReadingHistory(sourceId: manga.sourceId, mangaId: manga.id)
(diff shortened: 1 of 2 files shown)


zabroseric:
Example of UI
image

Note: I'm a React / NodeJS dev, so feel free to request corrections to better follow project/language standards.

iOS/Scenes/Library/LibraryViewController.swift
```diff
@@ -944,6 +944,25 @@ extension LibraryViewController {
...
+            let chapters = (try? await source?.getChapterList(manga: manga)) ?? []
+            DownloadManager.shared.download(chapters: chapters, manga: manga)
+        }
+    }
+
+    func downloadUnreadAction(manga: Manga) {
+        Task {
+            let source = SourceManager.shared.source(for: manga.sourceId)
+            let readingHistory = await CoreDataManager.shared.getReadingHistory(sourceId: manga.sourceId, mangaId: manga.id)
+            let chapters = ((try? await source?.getChapterList(manga: manga)) ?? []).filter {
```
Skittyblock, line 959: probably best to use `CoreDataManager.shared.getChapters` instead https://github.com/Aidoku/Aidoku/blob/d19878092bfe2168d95e7a234b743d55d320580e/Shared/Managers/CoreData/CoreDataManager%2BChapter.swift#L51

Skittyblock (changes requested):
might as well set the two download functions as `async` rather than spawning a task inside them and then instead spawn a task when you call them

zabroseric:
Awesome, thanks @Skittyblock for the info, especially considering it would just be quicker for
you to resolve it yourself. I've fixed what you mentioned, and it makes sense. It feels like the download functionality (all/unread) should be moved out of the controller, and into something like the DownloadManager. Thoughts?

iOS/Scenes/Library/LibraryViewController.swift
```diff
@@ -944,6 +944,19 @@ extension LibraryViewController {
...
 
+    func downloadAll(manga: Manga) async {
+        let chapters = await CoreDataManager.shared.getChapters(sourceId: manga.sourceId, mangaId: manga.id)
+        DownloadManager.shared.download(chapters: chapters, manga: manga)
+    }
+
+    func downloadUnread(manga: Manga) async {
+        let readingHistory = await CoreDataManager.shared.getReadingHistory(sourceId: manga.sourceId, mangaId: manga.id)
+        let chapters = await CoreDataManager.shared.getChapters(sourceId: manga.sourceId, mangaId: manga.id).filter {
+            readingHistory[$0.id] == nil || readingHistory[$0.id]?.page != -1
```
zabroseric, line 955: Ensure partially, but unread chapters are downloaded.

Skittyblock:
yeah might be good to add the functions to `DownloadManager` instead since I might add an option in the manga menu for downloading (like unread and ranges, etc.)

zabroseric:
Sweet, that change has been made. Yeah, I was thinking the same and played around with the idea below:

image

Skittyblock:
I would probably just do a child menu in the ellipsis menu rather than a popover but yeah

Skittyblock (approved):

zabroseric:
Sweet, your project, your call. :)

PR #209, opened 2023-04-02
Adding an optional "Read all chapters in this manga" setting

pandeynmn:
Adding an optional button in Manga Page enabled in the settings which allows for all the chapters to be marked as read in a single click instead of the painful process of selecting all, and then marking as read. 

Useful when converting library or mass marking your library. Can be turned off in settings anyway.

image
image
image


Skittyblock:
just add this back if you want it. I don't think we need a setting option for it, it doesn't make sense. also best not to add localization keys manually. https://github.com/Aidoku/Aidoku/blob/d635251eb2e5e301df7e347c33527b6594ac6429/iOS/UI/Manga/MangaViewController.swift#L281-L300

pandeynmn:
I added a setting option because I can see people accidentally clicking it when they want to click share or migrate.

Skittyblock:
yeah that's why it's best as a submenu so you need two taps, and you have an extra option to mark read or unread

pandeynmn:
but in the case where you are a new user, or just mass-reading the library, the three step process makes it harder if you have more than a couple you have to do. I see this more as a qol thing.

iOS/UI/Settings/SettingsViewController.swift
Skittyblock, line 1: revert changes on this file

iOS/Scenes/Manga/MangaViewController.swift
```diff
@@ -580,7 +602,7 @@ extension MangaViewController {
                 })
             }
         }
-
+
```
Skittyblock, line 605: don't add extra spaces

iOS/Scenes/Manga/MangaViewController.swift
```diff
@@ -540,7 +540,29 @@ extension MangaViewController {
 
     private func makeMenu() async -> [UIMenuElement] {
         var menus = [UIMenu]()
-        var actions = [
+        var actions: [UIMenuElement] = [
+            // mark as read & mark as unread submenu
+            UIMenu(title: NSLocalizedString("MARK_ALL", comment: ""), image: UIImage(systemName: "pencil.tip.crop.circle"),children: [
```
Skittyblock, line 545: remove the image for this I think, and add a space after the comma
```suggestion
            UIMenu(title: NSLocalizedString("MARK_ALL", comment: ""), image: nil, children: [
```

iOS/Scenes/Manga/MangaViewController.swift
```diff
@@ -540,7 +540,29 @@ extension MangaViewController {
...
+                // read chapters
+                UIAction(title: NSLocalizedString("READ", comment: "")) { _ in
+                    self.showLoadingIndicator()
+                    let chapters = [Chapter](self.viewModel.chapterList)
```
Skittyblock, line 549: why do you need to cast this? just store it

Skittyblock (changes requested):

iOS/Scenes/Manga/MangaViewController.swift
pandeynmn: removed casting, directly used in the function

iOS/Scenes/Manga/MangaViewController.swift
pandeynmn: removed the icon from the menu

Skittyblock (approved):

PR #222, opened 2023-04-19
Translation and other language support

arch-fan:
Made Spanish readme and link to both readmes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
 # Aidoku
+
+English | [Español](README_ES.md)
+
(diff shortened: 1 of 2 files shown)


Skittyblock:
Thanks for the PR, but I think I'm going to decide against merging this. The README doesn't really have important information that an automatic machine translator wouldn't be able to convey, and there isn't really much sense maintaining multiple translations of it.

If we had, say, a contributing guide instead, that would make much more sense to translate I think.

Issue #191, opened 2023-03-03
Add haptic feedback to the page slider at the bottom


ryansony:
Adding haptic feedback to the page slider at the bottom of the manga reader view would be pretty helpful in distinguishing page steps. An example of a similar implementation is the YouTube iOS app which has haptic feedback as you seek through video chapters on the timeline. I feel like I missed that here and would be a good addition.

PR #248, opened 2023-06-04
[slider] add haptic feedback when slider changes page

annuhdo:
# Context
Issue: https://github.com/Aidoku/Aidoku/issues/191

- Added selection haptic feedback when `ReaderViewController#displayPage` gets called.
- Stored `currentPageLabelVal` to `ReaderToolbarView.swift` so we can figure out if the user had slid to a page that differed from previous page. We only fire the haptic feedback when the page changes.

# Test
- Can't build this on physical device so I can't really test.
- Ran print statements which ran in the if block.

----

I'm new to iOS development so would appreciate any feedback! I'm also aware that the test plan is lacking so feel free to reject the PR but pls provide feedback on code if possible :)

### DIFF
--- a/iOS/UI/Reader/ReaderToolbarView.swift
+++ b/iOS/UI/Reader/ReaderToolbarView.swift
@@ -10,15 +10,10 @@ import UIKit
 class ReaderToolbarView: UIView {
     var currentPageValue: Int? {
         didSet {
-            if previousPageValue != currentPageValue {
-                previousPageValue = currentPageValue
+            if oldValue != currentPageValue {
+                let feedbackGenerator = UISelectionFeedbackGenerator()
+                feedbackGenerator.selectionChanged()
             }
-        }
-    }
-    var previousPageValue: Int? {
-        willSet {
-            let feedbackGenerator = UISelectionFeedbackGenerator()
-            feedbackGenerator.selectionChanged()
         }
     }
     var currentPage: Int? {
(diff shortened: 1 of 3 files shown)


ItsLame:
Just tested it on my iPhone. Rather than the haptic feedback kicking in when page changes, it instead continuously triggers haptic when dragging the slider.

annuhdo:
@ItsLame oof. Let me try again and test on a test project.

Not familiar with how to set this PR as planning changes but uh —planning changes— now.

annuhdo:
Will put up another PR

ItsLame:
@annuhdo I think you can just make updates to your current branch (`hapticslider`) and it'll show up here

ItsLame:
> Not familiar with how to set this PR as planning changes but uh —planning changes— now.

Also for this I usually just set the PR back to draft like so:
image

ItsLame:
@annuhdo just noticed, I think you forgot to re-open this PR?

iOS/UI/Reader/ReaderToolbarView.swift
```diff
@@ -8,7 +8,19 @@
...
+            if previousPageValue != currentPageValue {
+                previousPageValue = currentPageValue
+            }
+        }
+    }
+    var previousPageValue: Int? {
+        willSet {
+            let feedbackGenerator = UISelectionFeedbackGenerator()
+            feedbackGenerator.selectionChanged()
+        }
```
Skittyblock, line 22: you don't need two variables, just use `oldValue`
```suggestion
            if oldValue != currentPageValue {
                let feedbackGenerator = UISelectionFeedbackGenerator()
                feedbackGenerator.selectionChanged()
            }
        }
```

Skittyblock (changes requested):

PR #259, opened 2023-06-19
Feature crop borders

axiel7:
- Added to Manga reader only (idk if webtoons would benefit from this)
- Seems to work fine on white backgrounds and simple black backgrounds
- Added as a setting, off by default

Feel free to request any code changes!


axiel7:
It was lagging when fast scrolling pages so I used the same approach as downsampling and added a new processor

iOS/UI/Reader/Page/CropBordersProcessor.swift
```diff
@@ -0,0 +1,124 @@
...
+//
+
+import Foundation
+import Nuke
+
+#if os(iOS) || os(tvOS)
+import UIKit
+#else
+import CoreGraphics
+#endif
```
Skittyblock, line 15: you only need to import coregraphics
```suggestion
import CoreGraphics
```

Skittyblock:
seems to work pretty good, nice! the only thing I'd want is to have it refresh the pages after disabling/enabling the setting, as currently it only takes effect on pages that haven't been loaded yet.

as a side note, the same thing should be done for the downsample images option (but I forgot), but you don't need to include that in this pr.

iOS/UI/Reader/Page/ReaderPageView.swift
```diff
@@ -167,6 +172,13 @@ class ReaderPageView: UIView {
                         image = processedImage
                     }
                 }
+                if UserDefaults.standard.bool(forKey: "Reader.cropBorders") {
+                    let processor = CropBordersProcessor()
+                    let processedImage = processor.process(image)
+                    if let processedImage = processedImage {
```
Skittyblock, line 178: for future reference (I'll add this to the swiftlint rules in the future as I need to change it throughout the app)
```suggestion
                    if let processedImage {
```

Skittyblock (approved):

PR #265, opened 2023-06-30
Add swipe actions to manga chapters list

axiel7:
Opening this again because I forgot to create a branch just for this :)


Skittyblock:
still on the fence about whether or not to actually add this, but regardless it doesn't seem to work... marking a chapter unread/read with the swipe action doesn't animate properly and doesn't update the action afterwards.

axiel7:
Fixed the animation and state, if you decide to add it

PR #266, opened 2023-07-01
Add manga cover view

### LINKED ISSUE #143: Save manga cover image option

axiel7:
Added a simple view to visualize the manga cover in full size and options to save/share.
I'm not used to UIKit but I think everything is okay, feel free to request any code change.

Closes #143


Skittyblock (changes requested):
- use a "done" navigation bar button instead of a regular button
- the image already exists on the manga page, so there's no reason to load it again, just pass it into the initializer and reuse it
- I think it'd probably be better as a fullscreen sheet with a navigation bar title (like "Cover")
- the image view should be the same aspect ratio as the image so that the force touch shadow is correctly aligned
- the image view should also be inset a little, centered, and probably add a bit of corner radius to make it look better.

axiel7 (commented):
Done everything as requested (except passing image from the manga page because it's downsampled)

Skittyblock:
changes look perfect! I forgot to say that we should also dim the image on touch down though like swiftui would do by default for a button

axiel7:
> changes look perfect! I forgot to say that we should also dim the image on touch down though like swiftui would do by default for a button

I tried to implement it with a `UILongPressGestureRecognizer` but it seems that `UIContextMenuInteraction` overrides it.

It's really necessary? I don't think that's the default behaviour, when I long press a photo in the iOS Photos app the image doesn't dim

Skittyblock:
not long press, on touch down. in the photos app it doesn't do it since the photo directly expands into the view, but stuff like app icons dim when you press on them

Skittyblock (approved):
added a custom gesture that allows for canceling and doesn't prevent scrolling, but rest looks good thanks

PR #302, opened 2023-08-09
Pinned source feature

### LINKED ISSUE #201: Suggestion: Pin Sources
I would like an option to pin sources to the top of the list please. While I try to use the original TL sites, I usually default to only 3 for reading/discovering, so being able to find them easily would be really nice.

Thank you for all the hard work creating this app.

pandeynmn:
Ability to pin and re-order sources in a new section above the installed sources section (while remaining below the update source section).

Closes #201   (Closes 201 not 244)

https://i.imgur.com/R0k8wjo.mp4

### DIFF
--- a/Shared/Sources/SourceManager.swift
+++ b/Shared/Sources/SourceManager.swift
@@ -165,7 +165,6 @@ extension SourceManager {
             pinnedList.append(source.id)
             UserDefaults.standard.set(pinnedList, forKey: key)
         }
-        // Notify Observers
         NotificationCenter.default.post(name: Notification.Name("updateSourceList"), object: nil)
     }
 
@@ -177,7 +176,6 @@ extension SourceManager {
             pinnedList.remove(at: index)
             UserDefaults.standard.set(pinnedList, forKey: key)
         }
-        // Notify Observers
         NotificationCenter.default.post(name: Notification.Name("updateSourceList"), object: nil)
     }
 }
--- a/Shared/Sources/SourceManager.swift
+++ b/Shared/Sources/SourceManager.swift
@@ -146,7 +146,6 @@ extension SourceManager {
     }
 
     func remove(source: Source) {
-        self.unpin(source: source) // Remove source from pin list.
         try? FileManager.default.removeItem(at: source.url)
         sources.removeAll { $0.id == source.id }
         Task {
--- a/Shared/Sources/SourceManager.swift
+++ b/Shared/Sources/SourceManager.swift
@@ -146,6 +146,7 @@ extension SourceManager {
     }
 
     func remove(source: Source) {
+        self.unpin(source: source) // Remove source from pin list.
         try? FileManager.default.removeItem(at: source.url)
         sources.removeAll { $0.id == source.id }
         Task {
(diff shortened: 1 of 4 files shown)


Skittyblock:
functionality seems good, but
- swiftlint issues should be fixed
- localization keys can be added to the english localization
  - keys should just be the words, "PIN", "UNPIN", and "REORDER" I think

iOS/UI/Browse/BrowseViewController.swift
```diff
@@ -364,3 +470,35 @@ extension BrowseViewController: UISearchResultsUpdating {
         updateDataSource()
     }
 }
+
+// Updating view
+extension BrowseViewController {
+    func updateNavBar() {
```
Skittyblock, line 476:
```suggestion
extension BrowseViewController {
    func updateNavbar() {
```

iOS/UI/Browse/BrowseViewController.swift
```diff
@@ -187,14 +185,27 @@ class BrowseViewController: BaseTableViewController {
...
+        .none
+    }
+
+    func tableView(_ tableView: UITableView, shouldIndentWhileEditingRowAt indexPath: IndexPath) -> Bool {
+        false
+    }
+
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         if
-            sectionIdentifier(for: indexPath.section) == .installed,
```
Skittyblock, line 197: use a `case let` here to store the identifier then check that it's one of the two

iOS/UI/Browse/BrowseViewController.swift
```diff
@@ -232,20 +243,66 @@ extension BrowseViewController {
...
+                    }]), UIAction(
+                        title: NSLocalizedString("UNPIN", comment: ""),
+                        image: UIImage(systemName: "pin.slash")
+                    ) { _ in
+                        // Remove source from the pinned array, recreate the installed source list and update the table.
+                        SourceManager.shared.unpin(source: source)
+                        self.viewModel.loadPinnedSources()
+                        self.updateDataSource()
+                    }, UIAction(
+                        title: NSLocalizedString("UNINSTALL", comment: ""),
```
Skittyblock, line 292: this action is duplicated for both, I think you could simplify the code so that you don't have to re-instantiate the same action in both branches

iOS/UI/Browse/BrowseViewController.swift
```diff
@@ -232,20 +243,66 @@ extension BrowseViewController {
         point: CGPoint
     ) -> UIContextMenuConfiguration? {
         guard
-            sectionIdentifier(for: indexPath.section) == .installed,
+            sectionIdentifier(for: indexPath.section) == .installed || sectionIdentifier(for: indexPath.section) == .pinned,
```
Skittyblock, line 246: same here, use `case let`

iOS/UI/Browse/BrowseViewController.swift
```diff
@@ -254,14 +311,59 @@ extension BrowseViewController {
...
+                    if destinationIndex > sourceIndex {
+                        snapshot.insertItems([sourceItem], afterItem: destinationItem)
+                    } else {
+                        snapshot.insertItems([sourceItem], beforeItem: destinationItem)
+                    }
+                }
+            }
+            // Save the order and notify the observer to reload table.
+            apply(snapshot, animatingDifferences: false)
+            NotificationCenter.default.post(name: Notification.Name("updatePinnedList"), object: nil)
```
Skittyblock, line 359: I don't think you should use notifications here, they're generally not a good option. probably use a weak var for the parent as a delegate since SourceCellDataSource is only used in this view controller.

Shared/Sources/SourceManager.swift
```diff
@@ -156,6 +156,30 @@ extension SourceManager {
...
+
+    // Pin a source in browse tab.
+    func pin(source: Source) {
+        let key = "Browse.pinnedList"
+        var pinnedList = UserDefaults.standard.stringArray(forKey: key) ?? []
+        if !pinnedList.contains(source.id) {
+            pinnedList.append(source.id)
+            UserDefaults.standard.set(pinnedList, forKey: key)
+        }
+        // Notify Observers
```
Skittyblock, line 168:
```suggestion
```

Skittyblock (changes requested):

Skittyblock:
also you can add a disable cyclomatic_complexity command for `filterExternalSources`

Skittyblock (changes requested):

iOS/UI/Browse/BrowseViewController.swift
```diff
@@ -254,14 +314,63 @@ extension BrowseViewController {
...
+        var onReorder: ((NSDiffableDataSourceSnapshot<Section, SourceInfo2>) -> Void)?
+        // Let the rows in the Pinned section be reordered (used for reordering sources)
+        override func tableView(_ tableView: UITableView, canMoveRowAt indexPath: IndexPath) -> Bool {
+            if #available(iOS 15.0, *) {
+                return sectionIdentifier(for: indexPath.section) == .pinned
+            } else {
+                let section = indexPath.section
+                guard section >= 0 else { return false }
+                let sections = self.snapshot().sectionIdentifiers
+                return ((sections.count > section ? sections[section]: Section.installed) == Section.pinned)
```
Skittyblock, line 336:
```suggestion
                return (sections.count > section ? sections[section] : Section.installed) == Section.pinned
```

Skittyblock (approved):

PR #329, opened 2023-08-27
Migrate to Nuke 12

axiel7:
- The `ImageTaskDelegate` was removed so I had to modify how the image tasks where loaded
- ~~I'm not really sure about the changes on `ReaderPageView`~~
- I added a custom `DataLoader` to the shared `ImagePipeline` as it's recommended in the [docs](https://kean-docs.github.io/nuke/documentation/nuke/cache-layers#Aggressive-Disk-Cache) if using a custom `DataCache`

### DIFF
--- a/iOS/Old UI/Base/Manga/MangaCoverCell.swift
+++ b/iOS/Old UI/Base/Manga/MangaCoverCell.swift
@@ -240,21 +240,18 @@ class MangaCoverCell: UICollectionViewCell {
             processors: [DownsampleProcessor(width: bounds.width)]
         )
 
-        imageTask = ImagePipeline.shared.loadImage(
-            with: request,
-            completion: { [weak self] result in
-                guard let self = self else { return }
-                switch result {
-                case .success(let response):
-                    Task { @MainActor in
-                        UIView.transition(with: self.imageView, duration: 0.3, options: .transitionCrossDissolve) {
-                            self.imageView.image = response.image
-                        }
+        imageTask = ImagePipeline.shared.loadImage(with: request) { [weak self] result in
+            guard let self = self else { return }
+            switch result {
+            case .success(let response):
+                Task { @MainActor in
+                    UIView.transition(with: self.imageView, duration: 0.3, options: .transitionCrossDissolve) {
+                        self.imageView.image = response.image
                     }
-                case .failure:
-                    imageTask = nil
                 }
+            case .failure:
+                imageTask = nil
             }
-        )
+        }
     }
 }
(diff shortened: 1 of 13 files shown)


iOS/UI/Reader/Page/ReaderPageView.swift
```diff
@@ -136,16 +136,27 @@ class ReaderPageView: UIView {
...
+                    completion?(true)
+                case .failure:
+                    completion?(false)
+                }
+                progressView.isHidden = true
+            }
+        )
 
-        return success
+        return true
```
Skittyblock, line 159: this will break things since the function is supposed to wait until the image loads; if there's an async version of `loadImage`, you could use that, otherwise use a `withCheckedContinuation` and return with that

iOS/UI/Migration/MangaGridView.swift
```diff
@@ -14,29 +14,35 @@ struct MangaGridView: View {
     var coverUrl: URL?
 
     var body: some View {
-        LazyImage(url: coverUrl, resizingMode: .aspectFill)
+        LazyImage(url: coverUrl) { state in
+            if let image = state.image {
+                image
+                    .resizable()
+                    .aspectRatio(contentMode: .fill)
+            }
```
Skittyblock, line 22: probably good to have a fallback image here like we do in the UIKit things
```suggestion
            } else {
                Image("MangaPlaceholder")
            }
```

Skittyblock (changes requested):

iOS/UI/Reader/Page/ReaderPageView.swift
axiel7: There's indeed an async function but it returns a new `AsyncImageTask` that doesn't have a `state` variable so I'll use `withCheckedContinuation`. Thanks

iOS/Old UI/Base/Manga/MangaCoverCell.swift
```diff
@@ -240,23 +240,18 @@ class MangaCoverCell: UICollectionViewCell {
             processors: [DownsampleProcessor(width: bounds.width)]
         )
 
-        do {
-            let image = try await ImagePipeline.shared.image(for: request, delegate: self).image
-            Task { @MainActor in
-                UIView.transition(with: imageView, duration: 0.3, options: .transitionCrossDissolve) {
-                    self.imageView.image = image
+        imageTask = ImagePipeline.shared.loadImage(with: request) { [weak self] result in
+            guard let self = self else { return }
```
Skittyblock, line 244:
```suggestion
            guard let self else { return }
```

Skittyblock (approved):
lgtm, thanks!

PR #363, opened 2023-10-17
Added Komga tracker

Paolo7297:
I have added a komga tracker to sync progress with my komga instance.
I also have added a tracker subclass implementing host-user-password tracker.
It's my first time with swift, I hope my code is OK.

### DIFF
--- a/Shared/Tracking/Tracker.swift
+++ b/Shared/Tracking/Tracker.swift
@@ -32,7 +32,7 @@ protocol Tracker: AnyObject {
     var isLoggedIn: Bool { get }
 
     /// An array of sources that this tracker can be used with
-    var restrictSources: [String]? { get }
+    var restrictSources: [String] { get }
 
     /// Register a new tracked title.
     ///
@@ -106,7 +106,7 @@ extension Tracker {
     func option(for score: Int) -> String? {
         scoreOptions.first { $0.1 == score }?.0
     }
-    var restrictSources: [String]? { nil }
+    var restrictSources: [String] { [] }
 }
 
 extension Tracker {
--- a/Shared/Tracking/TrackerManager.swift
+++ b/Shared/Tracking/TrackerManager.swift
@@ -26,8 +26,8 @@ class TrackerManager {
     /// A boolean indicating if there is a tracker that is currently logged in.
     func hasAvailableTrackers(manga: Manga? = nil) -> Bool {
         trackers.contains {
-            if let restrictSources = $0.restrictSources, let manga = manga {
-                return $0.isLoggedIn && restrictSources.contains(manga.sourceId)
+            if !$0.restrictSources.isEmpty, let manga = manga {
+                return $0.isLoggedIn && $0.restrictSources.contains(manga.sourceId)
             } else {
                 return $0.isLoggedIn
             }
--- a/Shared/Tracking/Trackers/komga/KomgaTracker.swift
+++ b/Shared/Tracking/Trackers/komga/KomgaTracker.swift
@@ -22,7 +22,7 @@ class KomgaTracker: HostUserPassTracker {
 
     let supportedStatuses = TrackStatus.defaultStatuses
 
-    var restrictSources: [String]? { ["multi.komga"] }
+    var restrictSources: [String] { ["multi.komga"] }
 
     var scoreType: TrackScoreType = .tenPoint
 
--- a/iOS/Old UI/Manga/Tracking/TrackerListView.swift
+++ b/iOS/Old UI/Manga/Tracking/TrackerListView.swift
@@ -17,7 +17,7 @@ struct TrackerListView: View {
     var body: some View {
         VStack {
             ForEach(TrackerManager.shared.trackers, id: \.id) { tracker in
-                if tracker.isLoggedIn && (tracker.restrictSources ?? [manga.sourceId]).contains(manga.sourceId) {
+                if tracker.isLoggedIn && (tracker.restrictSources.isEmpty ? [manga.sourceId] : tracker.restrictSources).contains(manga.sourceId) {
                     if let item = CoreDataManager.shared.getTrack(
                         trackerId: tracker.id,
                         sourceId: manga.sourceId,


luke-barnett:
@Skittyblock if @Paolo7297 resolves the merge conflicts aee we able to get this merged in?

Skittyblock:
@luke-barnett it still needs a little work to restrict it to only be used with the komga source, but the reason I haven't fixed it up myself is because the plan was to just integrate komga into aidoku directly instead of having a source for it eventually

Paolo7297:
@Skittyblock If I had known earlier, I would have done it right away!
Is it enough for merge or you want to integrate komga directly?

Skittyblock:
My bad, if it's restricted to work with the Komga source only and cleaned up a bit I can give it a more thorough review and merge it for the next update.

Paolo7297:
I resolved the conflicts, but for restrict the usage to komga only I think I should make deep changes in TrackerManager. Is it worth @Skittyblock ?

Paolo7297:
@Skittyblock you can check if this approach is what you want. 
I added a new var `restrictSources` to `Tracker`, so i can show it only with defined sources (in this case `multi.manga`.

Shared/Tracking/Tracker.swift
```diff
@@ -31,6 +31,8 @@ protocol Tracker: AnyObject {
     /// A boolean indicating if the tracker is currently logged in.
     var isLoggedIn: Bool { get }
 
+    var restrictSources: [String]? { get }
```
Skittyblock, line 34: add some documentation here
```suggestion
    /// An array of sources that this tracker can be used with
    var restrictSources: [String]? { get }
```

but I also don't think it should be optional. default value should just be an empty array.

Shared/Tracking/Trackers/myanimelist/MyAnimeListTracker.swift
```diff
@@ -31,6 +31,8 @@ class MyAnimeListTracker: OAuthTracker {
 
     var oauthClient: OAuthClient { api.oauth }
 
+    var restrictSources: [String]? { nil }
```
Skittyblock, line 34: this should be defined as a default for trackers, under where `scoreOptions` is set in Tracker.swift, so you don't have to define it in all trackers, just trackers that are actually restricted to certain sources

iOS/Old UI/Settings/TrackersViewController.swift
```diff
@@ -173,7 +173,58 @@ extension TrackersViewController {
             }
             session.presentationContextProvider = self
             session.start()
+        } else if let tracker = tracker as? HostUserPassTracker {
+            let loginPopUp = UIAlertController(title: String(format: NSLocalizedString("%@_LOGIN_TRACKER", comment: ""), tracker.name),
```
Skittyblock, line 177: indentation for this should be like
```swift
let loginAlert = UIAlertController(
    title: ...,
    message: ...,
    ...
)
```

and say Alert instead of PopUp

iOS/Old UI/Settings/TrackersViewController.swift
```diff
@@ -173,7 +173,58 @@ extension TrackersViewController {
...
+                let username = loginPopUp.textFields![1].text!
+                let password = loginPopUp.textFields![2].text!
+
+                Task{ @MainActor in
+                    let loadingIndicator = UIActivityIndicatorView(style: .medium)
+                    loadingIndicator.startAnimating()
+                    let res = await tracker.login(host: hostname, user: username, pass: password)
+
+                    if res
+                    {
```
Skittyblock, line 193: put braces on the same line as the `if` and `else` statements

iOS/Old UI/Settings/TrackersViewController.swift
```diff
@@ -173,7 +173,58 @@ extension TrackersViewController {
...
+                                               message: NSLocalizedString("INSERT_CREDENTIALS", comment: ""), preferredStyle: .alert)
+
+            let cancelAction = UIAlertAction(title: NSLocalizedString("CANCEL", comment: ""), style: .cancel, handler: nil)
+            let loginAction = UIAlertAction(title: NSLocalizedString("SAVE", comment: ""), style: .default) { _ in
+
+                let hostname = loginPopUp.textFields![0].text!
+                let username = loginPopUp.textFields![1].text!
+                let password = loginPopUp.textFields![2].text!
+
+                Task{ @MainActor in
```
Skittyblock, line 187:
```suggestion
                Task { @MainActor in
```

iOS/Old UI/Settings/TrackersViewController.swift
```diff
@@ -173,7 +173,58 @@ extension TrackersViewController {
...
+                textField.placeholder = "Password"
+                textField.isSecureTextEntry = true
+            }
+
+            loginPopUp.addAction(cancelAction)
+            loginPopUp.addAction(loginAction)
+
+            present(loginPopUp, animated: true, completion: nil)
         }
+
```
Skittyblock, line 227:
```suggestion
```

Shared/Tracking/TrackerManager.swift
```diff
@@ -17,13 +17,15 @@ class TrackerManager {
...
+    let komga = KomgaTracker()
 
     /// An array of the available trackers.
-    lazy var trackers: [Tracker] = [anilist, myanimelist]
+    lazy var trackers: [Tracker] = [anilist, myanimelist, komga]
 
     /// A boolean indicating if there is a tracker that is currently logged in.
-    var hasAvailableTrackers: Bool {
-        trackers.contains { $0.isLoggedIn }
+    func hasAvailableTrackers (manga: Manga?) -> Bool {
```
Skittyblock, line 27: default to providing no manga, since that was the prior behavior, and so you don't have to specify nil every time
```suggestion
    func hasAvailableTrackers(manga: Manga? = nil) -> Bool {
```

iOS/UI/Library/LibraryViewController.swift
```diff
@@ -804,7 +804,7 @@ extension LibraryViewController {
             }
         }
         let trackingFilter: [UIAction]
-        if TrackerManager.shared.hasAvailableTrackers {
+        if TrackerManager.shared.hasAvailableTrackers(manga: nil) {
```
Skittyblock, line 807: shouldn't you provide the manga here? i'm a little confused why it's used in some places and not others

iOS/Old UI/Manga/Tracking/TrackerListView.swift
```diff
@@ -17,7 +17,7 @@ struct TrackerListView: View {
     var body: some View {
         VStack {
             ForEach(TrackerManager.shared.trackers, id: \.id) { tracker in
-                if tracker.isLoggedIn {
+                if tracker.isLoggedIn && (tracker.restrictSources == nil || tracker.restrictSources!.contains(manga.sourceId) ){
```
Skittyblock, line 20: avoid `!` if you can
```suggestion
                if tracker.isLoggedIn && (tracker.restrictSources ?? []).contains(manga.sourceId) {
```

Shared/Localization/en.lproj/Localizable.strings
```diff
@@ -362,6 +364,9 @@
 "PILLARBOX_ORIENTATION" = "Pillarbox Orientation";
 "PILLARBOX_ORIENTATION_INFO" = "Pillarbox Orientation determines which orientations pillarboxing should be enabled for.";
 "BOTH" = "Both";
+"%@_LOGIN_TRACKER" = "Login to your %@ instance.";
+"INSERT_CREDENTIALS" = "Insert your credentials.";
+"ERROR_%@" = "Error while login to %@.";
```
Skittyblock, line 369:
```suggestion
"TRACKER_LOGIN_%@" = "Log in to %@";
"ENTER_CREDENTIALS" = "Enter your credentials.";
"TRACKER_LOGIN_ERROR_%@" = "Error while logging in to %@.";
```

Shared/Tracking/Trackers/komga/KomgaTracker.swift
```diff
@@ -0,0 +1,407 @@
...
+//  Created by Paolo7297 on 07/10/23.
+//
+
+import Foundation
+
+#if os(OSX)
+    import AppKit
+#else
+    import UIKit
+#endif
```
Skittyblock, line 14:
```suggestion
#if os(macOS)
import AppKit
#else
import UIKit
#endif
```

Shared/Tracking/Trackers/komga/KomgaTracker.swift
```diff
@@ -0,0 +1,407 @@
...
+                finishReadDate: nil
+            )
+        } else
+        {
+            return TrackState()
+        }
+    }
+
+    func getUrl(trackId: String) -> URL? {
+        URL(string: getSerieWebUrl(id: trackId)!)
```
Skittyblock, line 74: don't use `!`

Shared/Tracking/Trackers/komga/KomgaTracker.swift
```diff
@@ -0,0 +1,407 @@
...
+    private func decodeDate(_ value: AniListDate?) -> Date? {
+        if let day = value?.day, let month = value?.month, let year = value?.year {
+            return Calendar.current.date(from: DateComponents(year: year, month: month, day: day))
+        }
+        return nil
+    }
+
+    func createRequest(request: URLRequest, completionBlock: @escaping (String?, HTTPURLResponse?, Error?) -> Void)
+    {
+        let semaphore = DispatchSemaphore(value: 0)
```
Skittyblock, line 264: why are you using semaphores..? please use async instead

Shared/Assets.xcassets/komga.imageset/komga.svg
Skittyblock, line 1: since this is a single size svg, you should be able to set it as a universal image set. there's no reason to have three of these same icons

Skittyblock (changes requested):

Shared/Tracking/Tracker.swift
Paolo7297: I thought nil could be more relevant for a non working filter, instead if this variable was meant for disabling a source, the empty array could be a better option.

Paolo7297:
Everything should be okay now.

Shared/Tracking/Tracker.swift
Skittyblock: when would you have it be nil instead of an empty array? the convention in swift is to not have optional arrays, it's to use an empty array in the same way as you would nil

Paolo7297: I'm not experienced in swift, sorry. Now it's an empty array

dyphire:
Any progress? Why it stagnated again.

Jamerrone:
> Any progress? Why it stagnated again.

Glad I am not the only one who desperately wants this hahaha. I assume he is just busy with the source refactor.

Skittyblock:
the code style needs a lot of changes so i was dragging my feet doing that myself, and since i was planning on writing komga straight into the app rather than as a source in 0.7 (or 0.7.1 or something), that would change the implementation again

PR #393, opened 2024-01-16
Fix double url string encoding in ios17

axiel7:
- Fixes mangadex search query and possibly other sources

It works but I'm still not sure, let me know if something can be improved.


Skittyblock (approved):
I'll test it before I merge before the next update, but looks fine to me

Shared/Wasm/Imports/WasmNet.swift
```diff
@@ -239,7 +239,18 @@ extension WasmNet {
     var send: (Int32) -> Void {
         { descriptor in
             guard let request = self.globalStore.requests[descriptor] else { return }
-            guard let url = URL(string: request.URL ?? "") else { return }
+            var url: URL?
```
Skittyblock, line 242:
```suggestion
            let url: URL?
```

PR #394, opened 2024-01-21
Download with Wi-Fi only

### LINKED ISSUE #348: Download with Wi-Fi only toggle option
Thinking of the feature that Tachiyomi also had of being able download only with Wi-Fi as a toggle like the one that is currently on library update
Only on Wifi option

urbaneewe:
Added the toggle in the settings to download with Wi-Fi only.
Also added the alerts if the user tries to download while not on Wi-Fi.

Closes #348


iOS/UI/Manga/MangaViewController.swift
```diff
@@ -444,6 +444,18 @@ class MangaViewController: BaseTableViewController {
         present(UIHostingController(rootView: SwiftUINavigationView(rootView: AnyView(migrateView))), animated: true)
     }
 
+    func showNoWifiAlert() {
+        let alertController = UIAlertController(
+            title: NSLocalizedString("NO_WIFI_ALERT_TITLE", comment: ""),
+            message: NSLocalizedString("NO_WIFI_ALERT_MESSAGE", comment: ""),
+            preferredStyle: .alert)
```
Skittyblock, line 451:
```suggestion
            preferredStyle: .alert
        )
```

iOS/UI/Manga/MangaViewController.swift
```diff
@@ -444,6 +444,18 @@ class MangaViewController: BaseTableViewController {
...
+
+        let okAction = UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .default)
+        alertController.addAction(okAction)
+
+        present(alertController, animated: true, completion: nil)
```
Skittyblock, line 456: I don't think you need the completion parameter since it defaults to nil
```suggestion
        present(alertController, animated: true)
```

iOS/UI/Manga/MangaViewController.swift
```diff
@@ -894,7 +911,13 @@ extension MangaViewController {
...
                     image: UIImage(systemName: "arrow.down.circle")
                 ) { _ in
-                    DownloadManager.shared.download(chapters: [chapter], manga: self.manga)
+                    if UserDefaults.standard.bool(forKey: "Library.downloadOnlyOnWifi") && Reachability.getConnectionType() == .wifi {
+                        DownloadManager.shared.download(chapters: [chapter], manga: self.manga)
+                    } else {
+                        self.showNoWifiAlert()
+                    }
+
+//                    DownloadManager.shared.download(chapters: [chapter], manga: self.manga)
```
Skittyblock, line 920:
```suggestion
```

iOS/UI/Library/LibraryViewController.swift
```diff
@@ -993,14 +993,23 @@ extension LibraryViewController {
...
-                Task {
-                    await DownloadManager.shared.downloadUnread(manga: manga.toManga())
+                if UserDefaults.standard.bool(forKey: "Library.downloadOnlyOnWifi") && Reachability.getConnectionType() == .wifi {
+                    Task {
+                        await DownloadManager.shared.downloadUnread(manga: manga.toManga())
+                    }
+                } else {
+                    self.showNoWifiAlert()
                 }
+
```
Skittyblock, line 1012:
```suggestion
```

Skittyblock (changes requested):
here are a few style changes.

I'll check out the functionality later, but I think it might be a good idea to add in the hanging downloads when disconnected from wifi before this is merged. it should be as simple as just checking if the setting is on and if wifi isn't connected whenever it tries to move on to a new download

urbaneewe:
@Skittyblock Hey Skitty. Couple of questions and some possible improvements before I make more commits.

- When there are multiple downloads in the queue and the wifi suddenly cuts do we just cancel all the downloads and show an alert?

- Besides canceling I have 2 suggestions: Storing the downloads or simply pausing them. I see some beginnings of the code for both suggestions but I'm not sure how you planned to integrate it. For now I can just cancel the downloads if you don't want these implemented

- Also I may have found a solution for the multiple declarations of the `showNoWifiAlert` function. I was thinking I can change the definitions of the download methods in the `DownloadManager` to include an `@escaping` closure that contains the alert. That way whenever the function is called we can pass whatever we want for the alert.

Skittyblock:
Ideally, downloads should be paused instead of cancelled. Then there should be a button to resume that will show an alert instead of actually resuming if you're not on wifi and you have the setting enabled.

I'm not sure exactly what you mean about adding a closure to `DownloadManager` methods. Do you mean like a completion handler? That isn't necessary since it's async. Do you mean to include the actual alert in the method and then pass the contents or action or whatever to it in the arguments? I'd prefer to not do that since it's a shared manager that should be platform agnostic (e.g. doesn't have any ui code like alerts).

iOS/UI/Library/LibraryViewController.swift
Skittyblock, line 1: there should be a way to decrease the amount of code duplication here since you're doing essentially the same thing like 4 time, but honestly it's whatever since I know this stuff will be rewritten eventually.

Skittyblock (changes requested):

iOS/UI/Library/LibraryViewController.swift
urbaneewe: Probably leave it like this for now since I don't know what your plans for rewriting are.
I've pulled out the alert creation logic in a separate extension.

One thing that comes to mind here would be to have an `enum` for different download actions (all, unread) and then we would only need one check of the settings with the `switch/case` doing the task of calling the right method

Aidoku.xcodeproj/project.pbxproj
```diff
@@ -2055,6 +2061,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
+
```
Skittyblock, line 2064:
```suggestion
```

Skittyblock (changes requested):
Alright the only issue I see is you can't tell easily when downloads are paused. I think I would, instead of doing a menu with clear and pause/resume buttons, just put the two buttons on the left side of the navbar, and then people can see if it's paused without having to open the menu. But if you have a better idea that'd be fine.

urbaneewe:
> Alright the only issue I see is you can't tell easily when downloads are paused. I think I would, instead of doing a menu with clear and pause/resume buttons, just put the two buttons on the left side of the navbar, and then people can see if it's paused without having to open the menu. But if you have a better idea that'd be fine.

Well, we had a discussion about this on Discord where you suggested the UIMenu because having two buttons one next to another looked ugly and we didn't want to do floating buttons because those are too Android like.

I don't really have an alternative suggestion here so I can do the buttons one next to another if you really want that

Shared/Extensions/UIViewController+Alerts.swift
```diff
@@ -0,0 +1,27 @@
...
+                let okAction = UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .default)
+                alertController.addAction(okAction)
+            } else {
+                for action in actions {
+                    alertController.addAction(action)
+                }
+            }
+
+            self.present(alertController, animated: true, completion: completion)
+        }
```
Skittyblock, line 26:
```suggestion
        let alertController = UIAlertController(title: title, message: message, preferredStyle: .alert)

        // If no actions are provided, add a default 'OK' action
        if actions.isEmpty {
            let okAction = UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .cancel)
            alertController.addAction(okAction)
        } else {
            for action in actions {
                alertController.addAction(action)
            }
        }

        self.present(alertController, animated: true, completion: completion)
    }
```

Shared/Extensions/UIViewController+Alerts.swift
Skittyblock, line 1: This shouldn't be under the Shared folder since it's only for iOS, not cross platform

Skittyblock (changes requested):
rest looks good to me

iOS/Extensions/UIViewController+Alerts.swift
```diff
@@ -0,0 +1,27 @@
...
+import Foundation
+import UIKit
+
+extension UIViewController {
+    func presentAlert(title: String, message: String, actions: [UIAlertAction] = [], completion: (() -> Void)? = nil) {
+        let alertController = UIAlertController(title: title, message: message, preferredStyle: .alert)
+
+        // If no actions are provided, add a default 'OK' action
+        if actions.isEmpty {
+            let okAction = UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .default)
```
Skittyblock, line 17:
```suggestion
            let okAction = UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .cancel)
```

PR #400, opened 2024-02-10
Manga updates implementation

axiel7:
TODO:
- [x] Decide a better place to put the Manga Updates screen
- [x] Refresh list when read state changes
- [x] Test if the updates are being created correctly
- [x] Read/unread state


Shared/Aidoku.xcdatamodeld/0.6.xcdatamodel/contents
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="21279" systemVersion="21G115" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="22522" systemVersion="23D60" minimumToolsVersion="Automatic" sourceLanguage="Swift" userDefinedModelVersionIdentifier="">
```
axiel7, line 2: I ended up modifying the 0.6 data model, let me know what should I do with that

iOS/Extensions/View.swift
```diff
@@ -26,3 +26,14 @@ extension View {
...
     }
 }
+
+extension View {
+    @ViewBuilder
+    func refreshableCompat(action: @Sendable @escaping () async -> Void) -> some View {
+        if #available(iOS 15.0, *) {
+            self.refreshable(action: action)
+        } else {
+            self
```
axiel7, line 36: iOS 14 won't have a swipe to refresh, but I think you don't care (?)

iOS/UI/Updates/MangaUpdatesView.swift
```diff
@@ -0,0 +1,188 @@
...
+            loadingTask = nil
+        }
+        loadingMore = true
+        entries = []
+        offset = 0
+        reachedEnd = false
+        await loadNewEntries()
+    }
+
+    private func makeRelativeDate(_ days: Int) -> String {
```
axiel7, line 163: I copied this method from History, I think we should put it in a *Date extensions* or smth

Shared/Managers/Manga/MangaManager.swift
```diff
@@ -253,12 +253,23 @@ extension MangaManager {
...
-                                CoreDataManager.shared.setChapters(
+                                let newChapters = CoreDataManager.shared.setChapters(
                                     chapters,
                                     sourceId: manga.sourceId,
                                     mangaId: manga.id,
                                     context: context
                                 )
+                                // update manga updates
+                                for chapter in newChapters
+                                where mangaObject.langFilter != nil ? chapter.lang == mangaObject.langFilter : true
```
axiel7, line 264: not sure if we should do this

Shared/Data/Database/Objects/MangaUpdateObject.swift
Skittyblock, line 1: this probably could just be generated automatically instead of manually i think, and then you can just add the id extension to the generated class

Shared/Managers/Manga/MangaManager.swift
Skittyblock: I think we should, this is good. people that have a chapter language filter set probably don't change it often, and probably only care about the chapters for that particular language anyways

iOS/Extensions/Dates.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  Dates.swift
+//  Aidoku (iOS)
+//
+//  Created by axiel7 on 13/02/2024.
+//
+
+import Foundation
+
+func makeRelativeDate(days: Int) -> String {
```
Skittyblock, line 10: would it make sense to put this as a static function on a `Date` extension or something? just having global functions makes me feel a little weird

iOS/UI/Common/BaseViewController.swift
```diff
@@ -7,6 +7,7 @@
 
 import UIKit
 import AsyncDisplayKit
+import SwiftUI
```
Skittyblock, line 10: I don't think you need to import SwiftUI here (?)
```suggestion
```

iOS/UI/Updates/MangaUpdateItemView.swift
```diff
@@ -0,0 +1,52 @@
...
+                    .lineLimit(2)
+
+                if let chapterTitle = item.chapter?.makeTitle() {
+                    Text(chapterTitle)
+                        .font(.footnote)
+                        .foregroundColor(.secondary)
+                        .lineLimit(2)
+                }
+            }
+        }// :HStack
```
Skittyblock, line 50: I think I prefer not having the comments indicating the closing bracket type...
```suggestion
        }
```

Skittyblock (changes requested):
CoreData changes should be put in a new 0.6.8 datamodel (we could change the name later if it doesn't make it into 0.6.8 though)

The only other thing is I really think we should group chapters from the same manga, particularly because pressing on one just opens the manga and marks all of them as 'viewed'. We could maybe do like an expandable thing where you click on the one item for the manga in the list and it expands to show all the chapters, or just combine them into one like the history cells and have a new chapter range and/or new chapter count.

Thanks for this whole implementation though, this will be really nice!

Shared/Data/Database/Objects/MangaUpdateObject.swift
axiel7: I changed it to be generated automatically and it seems CoreData always generates Strings as optional but whatever, I think it's a good trade off in this case

Shared/Data/Database/Objects/MangaUpdateObject.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  UpdateObject.swift
```
Skittyblock, line 2:
```suggestion
//  MangaUpdateObject.swift
```

Shared/Aidoku.xcdatamodeld/0.6.7.xcdatamodel/contents
Skittyblock, line 1: changes for this file should be reverted

Skittyblock (changes requested):

Skittyblock (approved):

Issue #268, opened 2023-07-06
Better handling of downloaded chapters


mossicrue:
Would be cool to add some features to add for a better handling downloaded chapters.

Suggested features cool to implement in a sections of settings tab:
- a button to delete all the downloaded chapter of all manga
- a statistics that shows the space used by the downloaded chapters

Will add other suggestion time by time

Daptonic:
I'd also really like the option to download the oldest unread chapters first, seems wierd that when downloading a new title, it starts at the end and not the beginning.

Skittyblock:
That issue in particular is a bug. It's supposed to download in lowest to highest chapter order.

CEBATOR:
"Remove All Downloads" button is already present in the app

Skittyblock:
done with download manager addition

PR #405, opened 2024-02-15
Reverse download order for all or unread chapters

CEBATOR:
### Background

Currently when downloading all chapters or unread chapters the download order is set from last to first.
This PR should reverse the order and instead download from first to last.
Having the order from first to last will allow users to start reading as soon as the first chapter is loaded.

### What has been done

1. Reversed order of chapters for `downloadAll` function
2. Reversed order of chapters for `downloadUnread` function

### How to test

1. Have at least a manga in the library
2. Long press the title from the library
3. From the menu select "Download" -> "All"
4. In the download Que observe that the download order is from first to last
5. Clear the queue
6. Mark a few chapters as read
7. Back from the library on long press select "Download" -> "Unread"
8. Make sure the order is from first unread to last

### Checklist

- [x] I didn't add new warnings
- [x] I didn't break localisations
- [x] I didn't break UI

### DIFF
--- a/Shared/Data/Downloads/DownloadManager.swift
+++ b/Shared/Data/Downloads/DownloadManager.swift
@@ -93,7 +93,7 @@ extension DownloadManager {
 
     func downloadAll(manga: Manga) async {
         let chapters = await CoreDataManager.shared.getChapters(sourceId: manga.sourceId, mangaId: manga.id)
-        download(chapters: chapters, manga: manga)
+        download(chapters: chapters.reversed(), manga: manga)
     }
 
     func downloadUnread(manga: Manga) async {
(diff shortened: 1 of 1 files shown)


CEBATOR:
Perhaps a pull request template is in order. This will streamline contributions to the project.

CEBATOR:
I did not notice this: https://github.com/Aidoku/Aidoku/issues/268#issuecomment-1627543013
Before opening the PR.

Skittyblock:
> Perhaps a pull request template is in order. This will streamline contributions to the project.

I agree. Probably an issue template too, maybe. I just haven't gotten around to looking into it.

Skittyblock (approved):
lgtm

PR #420, opened 2024-03-15
add feature copy while holding down

### LINKED ISSUE #404: Copy manga link in manga page
Holding down on the web view button should copy the link to your clipboard

ryugel:
Added copy to clipboard feature while web view button held

https://github.com/Aidoku/Aidoku/assets/69274926/c7a083c1-5fd7-4191-ac8d-4903fbfb6472

Closes #404


iOS/UI/Manga/MangaDetailHeaderView.swift
```diff
@@ -581,3 +586,11 @@ extension MangaDetailHeaderView: SizeChangeListenerDelegate {
         sizeChangeListener?.sizeChanged(bounds.size)
     }
 }
+
+extension UIButton {
```
Skittyblock, line 590: I don't think there's any reason to make this an extension on UIButtons, especially since it's only being used once. I would just move this code into the `configure()` function for the `safariButton`

iOS/UI/Manga/MangaViewController.swift
```diff
@@ -1160,6 +1160,21 @@ extension MangaViewController: MangaDetailHeaderViewDelegate {
         present(SFSafariViewController(url: url), animated: true)
     }
 
+    // copy manga link when holding down the web view button
+    func safariHeld() {
+        guard
+            let url = manga.url,
+            url.scheme == "https" || url.scheme == "http"
```
Skittyblock, line 1167: does the url scheme matter when copying it? it shouldn't be anything other than these two things in normal cases, but I don't think there's any reason to actually check it since if it were different it wouldn't really matter.

iOS/UI/Manga/MangaViewController.swift
```diff
@@ -1160,6 +1160,21 @@ extension MangaViewController: MangaDetailHeaderViewDelegate {
...
+        else { return }
+        UIPasteboard.general.string = url.absoluteString
+        let alert =
+        UIAlertController(title: NSLocalizedString("LINK_COPIED", comment: ""), message:
+                            NSLocalizedString("LINK_COPIED_TEXT",
+                                              comment: ""), preferredStyle: .alert)
+                    alert.addAction(UIAlertAction(title: "OK", style: .default, handler: nil))
+                    present(alert, animated: true, completion: nil)
```
Skittyblock, line 1175: indentation here is a little funky, and completion is nil by default
```suggestion
        let alert = UIAlertController(
            title: NSLocalizedString("LINK_COPIED", comment: ""),
            message: NSLocalizedString("LINK_COPIED_TEXT", comment: ""),
            preferredStyle: .alert
        )
        alert.addAction(UIAlertAction(title: "OK", style: .default, handler: nil))
        present(alert, animated: true)
```

Skittyblock (changes requested):
sorry for the delayed review!

iOS/UI/Manga/MangaDetailHeaderView.swift
ryugel: Thanks for the feedback. This piece of code was unnecessary, just changed it

iOS/UI/Manga/MangaViewController.swift
```diff
@@ -1160,6 +1160,21 @@ extension MangaViewController: MangaDetailHeaderViewDelegate {
...
+        else { return }
+        UIPasteboard.general.string = url.absoluteString
+        let alert = UIAlertController(
+                   title: NSLocalizedString("LINK_COPIED", comment: ""),
+                   message: NSLocalizedString("LINK_COPIED_TEXT", comment: ""),
+                   preferredStyle: .alert
+               )
+               alert.addAction(UIAlertAction(title: "OK", style: .default, handler: nil))
+               present(alert, animated: true)
+    }
```
Skittyblock, line 1176: indentation still looking a little off, also make sure to localize "OK" and use cancel style
```suggestion
        guard let url = manga.url else { return }
        UIPasteboard.general.string = url.absoluteString
        let alert = UIAlertController(
            title: NSLocalizedString("LINK_COPIED", comment: ""),
            message: NSLocalizedString("LINK_COPIED_TEXT", comment: ""),
            preferredStyle: .alert
        )
        alert.addAction(UIAlertAction(title: NSLocalizedString("OK", comment: ""), style: .cancel))
        present(alert, animated: true)
    }
```

Skittyblock (approved):

PR #425, opened 2024-03-25
Use "ellipsis.circle" for 3-dot menu

axiel7:
Most of Apple apps uses `ellipsis.circle` for the 3-dot icon menu. I personally think that it looks better:


Skittyblock:
It does seem like they've switched over, but I think I would personally do the gray background version since that's what music, podcasts, tv, etc. do and they're more in line with aidoku than notes or reminders.

IMG_647470A289F3-1

Skittyblock:
33e3a60e4dc0e149d5ad5506a512666850500163

PR #436, opened 2024-04-14
fix: empty chapter list when offline refreshing

### LINKED ISSUE #336: Downloaded chapters disappear if you refresh the screen while offline.
Ran into this yesterday, I didn't go offline, but my manga source did (which results in the same). This is a normal occurrence as I host my own manga and the server is only online when I turn my pc on. Anyway, as long as I don't refresh the app, my downloaded chapters keep working as expected, but if I refresh the page they disappear until the server is back on. This isn't a big issue, but it is annoying.

https://github.com/Aidoku/Aidoku/assets/16142485/bdb830ac-7d75-412c-be95-b9ff94a5439e

axiel7:
Closes #336

### DIFF
--- a/Shared/Extensions/Reachability.swift
+++ b/Shared/Extensions/Reachability.swift
@@ -17,7 +17,7 @@ enum NetworkDataType {
 
 class Reachability {
     static func getConnectionType() -> NetworkDataType {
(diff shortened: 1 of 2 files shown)


Skittyblock (approved):
hopefully there will be some actual error ui for when there are errors with the new source api, but this is better for now

Skittyblock (approved):

Issue #77, opened 2022-06-27
Light theme reader's background is black when the overlay is hidden


beer-psi:
https://user-images.githubusercontent.com/92439990/175904137-d772d1f0-6789-4909-a538-fb88f094d022.MOV

Skittyblock:
The reason I did that is because of the photos app, but should I leave it white? I know I do need to add more customization options for the reader view in the future though.

beerpiss:
> The reason I did that is because of the photos app, but should I leave it white? I know I do need to add more customization options for the reader view in the future though.

I think it should be white, or at least have an option to make it white, so it can match the manga page's background.

PR #441, opened 2024-05-11
feat: adds page background color option

axeII:
This PR introduces settings for page reader background color. 

Users can choose from three options: system theme (dark or light), light, or dark. The default value is dark, respecting the original default dark background but users can change it anytime. Additionally, English translation is included. 

The aim of this PR is to address the problem described in: [Issue 77](https://github.com/Aidoku/Aidoku/issues/77).

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -263,6 +263,10 @@
 "PAGED" = "Paged";
 "PAGE_LAYOUT" = "Page Layout";
 "DOUBLE_PAGE" = "Double Page";
+"PAGE_BG_COLOR" = "Page Background Color";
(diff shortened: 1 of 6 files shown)


Skittyblock (changes requested):
Could you turn this into a more general setting for both the paged and webtoon reader (e.g. the webtoon reader shows black bars if you enable pillarboxing, but there should be an option to have it be white)? 

Also, probably just say "Black" and "White" instead of "Dark" and "Light" since it's the literal background color.

axeII:
I relocated the setting option from Page Settings to General and renamed the choices to "White" and "Black". 
Please review this PR once more.

Skittyblock (approved):
lgtm, thanks for the contribution!

PR #448, opened 2024-05-26
Improve image loading for downloaded images

### LINKED ISSUE #359: Lag when reading downloaded chapters
If i download a chapter the swiping is stopping and not flawless. It is only flawless if you read it without download.

 https://github.com/Aidoku/Aidoku/assets/146199251/290f5745-7ad6-4b29-9753-5cf4a5ab4f22

kyleerhabor:
I've modified the page fetching behavior for image loading to skip reading the contents of downloaded images and creating a base64 string out of it.

I'm not familiar with most of the codebase, but from what I've browsed, it seems Aidoku uses base64 strings to encode images from sources via Wasm. This is used during the downloading process, but effectively unused when displaying downloaded pages.

The inspiration for this change came out of Aidoku's handling of large chapters. Because base64 is effectively compressed data, it doesn't serve well when many large images are stored in memory. In my experience, the memory usage causes iOS to kill the app past 2 GB (iPhone 12), as well as create a long delay prior to loading images.

Closes #359

### DIFF
--- a/Shared/Data/Downloads/DownloadManager.swift
+++ b/Shared/Data/Downloads/DownloadManager.swift
@@ -66,6 +66,19 @@ class DownloadManager {
         return pages.sorted { $0.index < $1.index }
     }
 
+    func getDownloadedPagesWithoutContents(for chapter: Chapter) -> [Page] {
+        cache.directory(for: chapter).contents
+            .compactMap { url in
+                Page(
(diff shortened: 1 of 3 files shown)


Skittyblock (approved):
Good work. This fixes the all lag that comes from reading large downloaded chapters from what I can see. Thank you!

PR #457, opened 2024-06-25
Add warning message when chapters are being skipped

### LINKED ISSUE #212: Show warning when skipping chapters
Show a warning if the next chapter you are about to read is a higher chapter number than expected.

Example from TachiJ2K:
TachiyomiJ2K

kanshres:
This closes #212 

A new UI label is added and displays when a chapter difference bigger than 1 is detected after going to a new chapter. Let me know if the text or styling should be changed. Maybe a warning icon should be displayed?

image

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -136,6 +136,7 @@
 "CURRENT_COLON" = "Current:";
 "NO_NEXT_CHAPTER" = "There is no next chapter";
 "NO_PREVIOUS_CHAPTER" = "There is no previous chapter";
+"SKIPPING_CHAPTERS" = "WARNING: Chapters are being skipped. Either the source is missing them or they are filtered out.";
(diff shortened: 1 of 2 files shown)


Skittyblock (changes requested):
We should make it more like tachiyomi; have the number of chapters being skipped, use the `exclamationmark.triangle.fill` sf symbol with a yellow color (like the trackers page), and put it in between the finished and next groups (since we're skipping chapters between them).

You can probably have the label just in between the other labels in the stack view, and make sure the width is limited so it's word wrapped properly.

kanshres:
This is how it looks now:
image

Tachiyomi for comparison:
image

Skittyblock:
wow thanks for making it just like I said... but now that I see it I feel like it's not quite right. maybe the warning icon should be above the text instead of next to it? you can probably leave it as-is and I'll take a look at it myself before I push the next update so we can get this merged before then, thanks

Skittyblock (approved):

PR #459, opened 2024-07-02
Add action for clearing history excluding mangas in library

### LINKED ISSUE #210: Suggestion: Option to Clean History except Library
Can we please have a way to delete history except for Mangas in our Library? It'll be a good way to clean up our account

kanshres:
This closes #210 

On the history tab, I added a new action for clearing all history outside of manga in the user's library.
image

Should this new action be added to Settings on "Clear Read History"? Also, let me know if the string or anything else should be changed.

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -17,6 +17,7 @@
 "REMOVE" = "Remove";
 "RENAME" = "Rename";
 "CLEAR" = "Clear";
+"CLEAR_EXCLUDING_LIBRARY" = "Clear Excluding Library";
 "PAUSE" = "Pause";
 "RESUME" = "Resume";
 "LOADING_ELLIPSIS" = "Loading…";
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -17,7 +17,6 @@
 "REMOVE" = "Remove";
 "RENAME" = "Rename";
 "CLEAR" = "Clear";
-"CLEAR_EXCLUDING_LIBRARY" = "Clear Excluding Library";
 "PAUSE" = "Pause";
 "RESUME" = "Resume";
 "LOADING_ELLIPSIS" = "Loading…";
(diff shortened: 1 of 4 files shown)


Skittyblock:
I think this is probably okay to have temporarily, but my actual plan was to have a page in settings that lists all the manga with history that aren't added to your library, and allows you to selectively delete them, or open them if you still have the source installed

kanshres:
> I think this is probably okay to have temporarily, but my actual plan was to have a page in settings that lists all the manga with history that aren't added to your library, and allows you to selectively delete them, or open them if you still have the source installed

Ah ok. Would that be a change for 0.7?

Skittyblock:
likely wont be in the initial 0.7 update since my goal is just to rewrite the source api and anything surrounding that which needs to be, but it would come after 0.7

kanshres:
Feel free to close this PR if you decide it's not needed

Skittyblock (changes requested):
Actually yeah I think a better place for this would be in the advanced section in the settings tab, under the "Clear Reading History" option

kanshres:
> Actually yeah I think a better place for this would be in the advanced section in the settings tab, under the "Clear Reading History" option

I've made the changes. Should the history be reloaded after clearing? I noticed it's done so in HistoryViewController but not in SettingsViewController

Skittyblock:
> Should the history be reloaded after clearing? I noticed it's done so in HistoryViewController but not in SettingsViewController

Is there a notification key for it? I can't remember. It probably should be, but depends on what you see while testing. If you open the history tab, then settings and clear history, then go back to the history tab, does it still display stuff?
I would assume it also needs to update all the badges in library as well.

kanshres:
> > Should the history be reloaded after clearing? I noticed it's done so in HistoryViewController but not in SettingsViewController
> 
> Is there a notification key for it? I can't remember. It probably should be, but depends on what you see while testing. If you open the history tab, then settings and clear history, then go back to the history tab, does it still display stuff? I would assume it also needs to update all the badges in library as well.

Yeah when I do the normal history clear, the items in the history tab are still there

Skittyblock (changes requested):
so yeah, add the notification sending to refresh the history tab and this should be good

kanshres:
I added the notification to normal clear and the new clear without library items in Settings

Skittyblock (approved):
lgtm, i'll get it in before the next update

PR #472, opened 2024-08-31
Add Remove All Downloads and Remove Read Downloads to Library view

NightQuest:
This also replaces "Remove All Downloads" in the MangaViewController with a sub-menu.


Skittyblock:
I'm on the fence about this. I feel like removing read downloads shouldn't be a common desirable action because of the "delete download after reading" option now. Leaning towards not having this.

I definitely don't think it would need to be added to the library manga cell force touch menu though.

NightQuest:
yeah that's fair tbh, I had added this to my personal fork a few months ago and forgot to PR until now.

I will say that it makes for easy cleanup.. if not this, there should be some way to easily do it somewhere.

Skittyblock:
I think in the future a downloads page in settings for managing downloaded chapters would be beneficial, since you can download chapters for manga you don't even have in your library, but it might be hard to keep track of them.

PR #474, opened 2024-08-31
Mark all as read/unread can now be accessed within the library directly

NightQuest:
This adds a "Mark All" sub-menu to the Library view's right click menu


Skittyblock (approved):
I guess we can have this too

PR #481, opened 2024-10-01
Support downloading multiple manga from selection

### LINKED ISSUE #467: Downloading from library multi-select doesn't work
When you select multiple mangas and pull up the push/hold menu. The download unread doesn't work.

kyleerhabor:
Closes #467.

A potential issue I see in this is mass-downloading a collection of manga from a source; however, there are some sources where this may be desirable (for example, [Komga](https://komga.org/)).


Skittyblock (changes requested):
I think while we're at it we should probably disable the other actions (manga info, edit categories, share) when more than one manga is selected, since they only work with a single manga.

you can ignore the swiftlint violates for the code you didn't add

kyleerhabor:
@Skittyblock I've made the changes locally, but think actions like removing from a category or the library make sense to support (it's fairly trivial to implement). Should I include them, or disable them as well?

Skittyblock:
if you would like to implement them, feel free. just make sure only actions that support multiple manga are shown when multiple are selected

kyleerhabor:
@Skittyblock I've added the relevant changes.

Skittyblock (approved):

PR #534, opened 2025-03-13
The changes I've made:

deep60:
1. Only attempt to load chapters from the database if the in-memory chapter list is empty
2. The chapter list is preserved in memory
3. The database is not updated with an empty chapter list
4.  The existing chapter objects in CoreData remain intact
Screenshot 2025-03-13 at 14 23 17

### DIFF
--- a/iOS/UI/Library/LibraryViewController.swift
+++ b/iOS/UI/Library/LibraryViewController.swift
@@ -960,6 +960,7 @@ extension LibraryViewController {
         return manga[path.row]
     }
 
+    // swiftlint:disable:next cyclomatic_complexity
     func collectionView(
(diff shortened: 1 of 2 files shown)


Skittyblock:
i mean... if the chapter list refresh fails, does it really make sense to load from db? if the refresh fails, it doesn't update the db anyways, and it's likely that the chapter pages would fail to load if the chapter list failed to load. what it *should* do is actually show an error, but there isn't ui for that yet. i feel like the only situation this would be useful in is to avoid downloaded chapters from disappearing when a source dies, but that's a separate issue and could be resolved in a more direct way rather than just one edge case. am i wrong though?

Skittyblock:
closing due to no response

PR #540, opened 2025-03-28
Mark skipped duplicate chapters as read

### LINKED ISSUE #517: Mark duplicate chapter skipped as read
There are plenty mangas with duplicate chapters.

And the skipping feature doesn’t mark them read.

This results in finding the chapter where we left of instead of just going back to it with a click, which doesn’t work because it’ll just go to the first chapter skipped since it’s **unread**

kanshres:
This fixes #517

Each chapter skipped in `getNextChapter()` will be appended to `chaptersToMark`. Once a chapter is finished, it and any skipped chapters are marked as read.

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -335,6 +335,7 @@
 "RENAME_CATEGORY_INFO" = "Enter a new name for this category.";
 "RENAME_CATEGORY_FAIL" = "Failed to Rename Category";
 "SKIP_DUPLICATE_CHAPTERS" = "Skip Duplicate Chapters";
+"MARK_SKIPPED_CHAPTERS" = "Mark Skipped Chapters as Read";
 "RENAME_CATEGORY" = "Rename Category";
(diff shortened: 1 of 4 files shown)


Skittyblock (changes requested):
this should be configurable setting in the settings tab under the skip duplicate chapters settings (that i guess depends on that setting also being enabled)

what i was wondering for this is if it would be better to automatically mark all duplicate chapters after reading a chapter, regardless of whether it was skipped or not? e.g. if there are two chapter 1s, reading one of them would mark both. i think that might be a better solution in the future, but this would be fine for now

kanshres:
> this should be configurable setting in the settings tab under the skip duplicate chapters settings (that i guess depends on that setting also being enabled)
> 
> what i was wondering for this is if it would be better to automatically mark all duplicate chapters after reading a chapter, regardless of whether it was skipped or not? e.g. if there are two chapter 1s, reading one of them would mark both. i think that might be a better solution in the future, but this would be fine for now

Yeah, I agree that would be better. getNextChapter() and getPreviousChapter() already search through the chapter list, so the function could be modified to automatically mark chapters it finds to be duplicates.

Skittyblock (approved):

PR #545, opened 2025-03-30
feat: save images to album

aphronyx:
## Related Issues

- Closes #537

## Changes

- Added support for saving images to the app album

## Screenshots

| Cover| Page | Album |
| :-: | :-: | :-: |
| cover | page | album |

## Notes

**Please let me know if there are any issues with this PR. Thanks for taking the time to review!**

PR #548, opened 2025-03-31
feat: check photo library access

aphronyx:
## Related Issues

- #545

## Changes

- Added support for opening settings if photo library access was denied

## Screenshots

| Before | After |
| :-: | :-: |
| <video src="https://github.com/user-attachments/assets/4998b187-881e-463c-8c52-5178f9f47cdc"> | <video src="https://github.com/user-attachments/assets/815d6678-8bde-421c-9836-4a3bbf14922a"> |

## Notes

### DIFF
--- a/iOS/Extensions/UIImage.swift
+++ b/iOS/Extensions/UIImage.swift
@@ -16,10 +16,16 @@ extension UIImage {
         return CGSize(width: pageSize.width, height: scaledHeight)
     }
 
-    func saveToAlbum(_ name: String? = nil) {
+    func saveToAlbum(_ name: String? = nil, viewController: BaseViewController) {
         let status = PHPhotoLibrary.authorizationStatus(for: .readWrite)
         if status == .restricted || status == .denied {
-            ()
+            viewController.confirmAction(
+                title: NSLocalizedString("ENABLE_PERMISSION", comment: ""),
+                message: NSLocalizedString("PHOTOS_ACCESS_DENIED_TEXT", comment: ""),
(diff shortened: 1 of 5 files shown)


iOS/Extensions/UIImage.swift
```diff
@@ -16,8 +16,23 @@ extension UIImage {
...
+            viewController.confirmAction(
+                title: NSLocalizedString("ENABLE_PERMISSION", comment: ""),
+                message: NSLocalizedString("PHOTOS_ACCESS_DENIED_TEXT", comment: ""),
+                continueActionName: NSLocalizedString("SETTINGS", comment: "")
+            ) {
+                if let settings = URL(string: UIApplication.openSettingsURLString) {
+                    UIApplication.shared.open(settings)
+                }
+            }
+        }
```
Skittyblock, line 31: should this just stop here if there’s no permission? the code after this would fail anyways, right?
```suggestion
        guard status != .restricted && status != .denied else {
            viewController.confirmAction(
                title: NSLocalizedString("ENABLE_PERMISSION", comment: ""),
                message: NSLocalizedString("PHOTOS_ACCESS_DENIED_TEXT", comment: ""),
                continueActionName: NSLocalizedString("SETTINGS", comment: "")
            ) {
                if let settings = URL(string: UIApplication.openSettingsURLString) {
                    UIApplication.shared.open(settings)
                }
            }
            return
        }
```

aphronyx: yes

PR #551, opened 2025-04-01
Modify marking duplicate chapters option

### LINKED ISSUE #547: Mark any duplicate chapters as read after reading one
optional setting. not necessarily just after skipping them; can probably replace that setting

explanation: https://github.com/Aidoku/Aidoku/pull/540#pullrequestreview-2727412401

kanshres:
Closes #547 

- Separated "Mark Duplicated Chapters" from requiring "Skip Duplicated Chapters" to be on
- changed getPrevChapter() to mark duplicate chapters from before it
- changed getNextChapter() to handle the two different settings:
  - If not skipping dupes but marking, return the next chapter in the list and mark dupes as read
  - If skipping but not marking, return the first non-duplicate chapter and keep dupes unread
  - If neither, return the next chapter in the list in first iteration
  - If both, return the first non-duplicate chapter and mark dupes

### DIFF
--- a/iOS/UI/Reader/ReaderViewController.swift
+++ b/iOS/UI/Reader/ReaderViewController.swift
@@ -391,7 +391,10 @@ extension ReaderViewController: ReaderHoldingDelegate {
 
         while index >= 0 {
             let new = chapterList[index]
-            let isDuplicate = new.chapterNum == chapter.chapterNum && new.volumeNum == chapter.volumeNum
+            let isDuplicate =
+                new.chapterNum == chapter.chapterNum
+                && new.volumeNum == chapter.volumeNum
+                && (!(new.chapterNum == nil && new.volumeNum == nil) || new.title == chapter.title)
 
             if nextChapterInList == nil {
                 nextChapterInList = new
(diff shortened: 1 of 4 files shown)


iOS/UI/Reader/ReaderViewController.swift
```diff
@@ -382,17 +382,28 @@ extension ReaderViewController: ReaderHoldingDelegate {
...
+        let markDuplicates = UserDefaults.standard.bool(forKey: "Reader.markDuplicateChapters")
+
         index -= 1
+        var nextChapterInList: Chapter?
+
         while index >= 0 {
             let new = chapterList[index]
-            if !skipDuplicates || (new.chapterNum != chapter.chapterNum || new.volumeNum != chapter.volumeNum) {
-                return new
+            let isDuplicate = new.chapterNum == chapter.chapterNum && new.volumeNum == chapter.volumeNum
```
Skittyblock, line 394: i think this will mark chapters that have no chapter or volume number as duplicates, but they probably shouldn’t be

PR #559, opened 2025-04-11
feat(reader): lock orientation

### LINKED ISSUE #48: Dedicated rotation lock toggle
Would be nice to have a rotation lock toggle in the reader settings so that the lock only applies to the reader.

aphronyx:
## Related Issue

- Closes #48

## Changes

- Added support for locking reader orientation 

## Screenshots

https://github.com/user-attachments/assets/b548dfd1-3d7a-4460-badb-e5b5b87afdde

## Notes


Skittyblock (approved):
thanks! i forgot you could do it like this

Issue #139, opened 2023-01-06
Previous read marks chapter read


robkooper:
Using version 0.6 when long pressing a chapter and selecting marking previous read, the chapter itself is marked read as well. With the name I was expecting only previous to be read and chapter itself not.

Skittyblock:
honestly yeah this was intended (see #10), but now I've kinda changed my mind and should probably change it.

Issue #10, opened 2022-03-26
Mark previous as read/unread marks current as read/unread


Groovin-Dev:
Selecting "Mark previous read/unread" also marks the selected chapter as read.

Skittyblock:
I discussed this and decided that the current functionality makes sense.

PR #560, opened 2025-04-11
Consistent sorting and fixes "Mark previous as read/unread"

xaskii:
Makes sure sorting in any direction/method lets you mark "previous" chapters rather than chapters below the selected one. Also avoids marking the current chapter as read/unread when using "Mark previous".

- For example, if you select chapter 10, regardless of sorting direction, chapters 0-9 will be marked.

Closes: https://github.com/Aidoku/Aidoku/issues/139

## New Behaviour
- Excludes the selected chapter when using "Mark previous"
- "Mark previous" now marks previous chapters  in terms of upload date, chapter number, and source-order.

- Sorting descending after the changes:
	- Upload date: latest to earliest
	- Chapter: latest to earliest
	- Source order: latest to earliest (same as before)
	
## Other
- Changed `MangaViewModel.getOrderedChapterList` to avoid breaking the reader.

Related: https://github.com/Aidoku/Aidoku/issues/10

### DIFF
--- a/iOS/UI/Manga/MangaViewController.swift
+++ b/iOS/UI/Manga/MangaViewController.swift
@@ -1037,14 +1037,19 @@ extension MangaViewController {
 
     /// Returns a "Mark Previous" submenu for the chapter cell at the specified index path.
     private func markPreviousSubmenu(at indexPath: IndexPath) -> UIMenu {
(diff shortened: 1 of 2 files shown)


Skittyblock (approved):
lgtm, thanks!

Issue #79, opened 2022-06-28
Shake to undo (e.g. removing manga from library)


beerpiss:

aphronyx:
- Library
  - [x] Remove from Library (#577)
  - [x] Remove from Category (#571)

PR #571, opened 2025-04-21
feat(library): undo remove from category

aphronyx:
## Related Issues

- #79

## Changes

- Added support for undoing and redoing removing manga from category

## Screenshots/Screen Recordings

https://github.com/user-attachments/assets/61fea331-9720-46db-b7a8-76689faf8087

## Notes

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -1,9 +1,9 @@
-/* 
+/*
   Localizable.strings
   Aidoku
 
   Created by Skitty on 4/23/22.
-  
+
 */
 
 // General
@@ -55,6 +55,8 @@
 "LATEST_CHAPTER" = "Latest Chapter";
 "DATE_ADDED" = "Date Added";
 "COVER" = "Cover";
+"REMOVING_(ONE)_MANGA_FROM_CATEGORY_%@" = "Removing Manga from Category %@";
+"REMOVING_%i_MANGA_FROM_CATEGORY_%@" = "Removing %i Manga from Category %@";
 
 // Categories
 "ALL" = "All";
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -55,8 +55,8 @@
 "LATEST_CHAPTER" = "Latest Chapter";
 "DATE_ADDED" = "Date Added";
 "COVER" = "Cover";
-"REMOVING_(ONE)_MANGA_FROM_CATEGORY_%@" = "Removing Manga from Category %@";
-"REMOVING_%i_MANGA_FROM_CATEGORY_%@" = "Removing %i Manga from Category %@";
+"REMOVING_(ONE)_ITEM_FROM_CATEGORY_%@" = "Removing Item from Category %@";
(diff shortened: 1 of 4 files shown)


aphronyx:
re-adding manga to library is quite complicated, so i only push category related first

iOS/UI/Library/LibraryViewController.swift
```diff
@@ -1162,3 +1158,43 @@ extension LibraryViewController: UISearchResultsUpdating {
...
+// MARK: - Undoable Methods
+extension LibraryViewController {
+    @discardableResult
+    func removeFromCategory(mangaInfo: [MangaInfo]) -> Task<Void, Never> {
+        let mangaCount = mangaInfo.count
+        let actionName =
+            mangaCount > 1
+            ? String(
+                format: NSLocalizedString("REMOVING_%i_MANGA_FROM_CATEGORY_%@", comment: ""),
+                mangaCount, viewModel.currentCategory!)
```
Skittyblock, line 1171: the force unwrap makes me a little uncomfortable. it *should* exist, but there could be some condition in which it ends up being `nil`.... to play it safe i'd just add
```swift
guard let currentCategory = viewModel.currentCategory else { return }
```
at the top of the function

aphronyx: which is better?

```swift
func removeFromCategory(mangaInfo: [MangaInfo]) -> Task<Void, Never> {
    guard let currentCategory = viewModel.currentCategory else { return Task {} }
    // …
}
```

```swift
func removeFromCategory(mangaInfo: [MangaInfo]) -> Task<Void, Never>? {
    guard let currentCategory = viewModel.currentCategory else { return nil }
    // …
}
```

Skittyblock: the latter I guess

Shared/Localization/en.lproj/Localizable.strings
```diff
@@ -55,6 +55,8 @@
 "LATEST_CHAPTER" = "Latest Chapter";
 "DATE_ADDED" = "Date Added";
 "COVER" = "Cover";
+"REMOVING_(ONE)_MANGA_FROM_CATEGORY_%@" = "Removing Manga from Category %@";
+"REMOVING_%i_MANGA_FROM_CATEGORY_%@" = "Removing %i Manga from Category %@";
```
Skittyblock, line 59: actually i think i changed my mind on these. since aidoku isn't exactly strictly a manga reader anymore, i think it'd be better to go with a more neutral "item" or "title" instead of "manga". i know it kinda conflicts with some of the other localized strings, but i think it'd be better for the future.
```suggestion
"REMOVING_(ONE)_ITEM_FROM_CATEGORY_%@" = "Removing Item from Category %@";
"REMOVING_%i_ITEMS_FROM_CATEGORY_%@" = "Removing %i Items from Category %@";
```

aphronyx: done

aphronyx:
undoing removing from library is done implemented, should i push to this pr or create another pr?

Skittyblock (approved):

Skittyblock:
just put it in another pr i guess; i can merge this one now

PR #584, opened 2025-05-17
Move some build settings to a build settings file

nschumac:
It was annoying me that when you want to compile and sideload it yourself. You'd need to change strings in the actual files to get iCloud sync etc. to work.

Something like this should be make it easier. 

You only need to edit one configuration file (change bundle suffix name for example) and remove the '-DCANONICAL_BUILD' compile flag to make it run and work properly on your device.


macOS/Aidoku-MACOS.xcconfig
```diff
@@ -0,0 +1,14 @@
...
+//  Aidoku-MACOS.xcconfig
+//  Aidoku
+//
+//  Created by Nikolai Schumacher on 17.05.25.
+//
+
+#include "../Shared/Aidoku.xcconfig"
+
+
+APP_ID_PREFIX               = xyz.skitty
```
Skittyblock, line 11: this should also be changed, not that the macOS app is used. i just forgot.
```suggestion
APP_ID_PREFIX               = app.aidoku
```

Aidoku.xcodeproj/project.pbxproj
```diff
@@ -2113,7 +2132,8 @@
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = iOS/Info.plist;
-				INFOPLIST_KEY_CFBundleDisplayName = "$(PRODUCT_NAME)";
+				INFOPLIST_KEY_CFBundleDisplayName = Aidoku;
```
Skittyblock, line 2135: does this remove the localization for e.g. the japanese app name?

Skittyblock (changes requested):

iOS/AppDelegate.swift
```diff
@@ -10,8 +10,13 @@ import Nuke
...
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
-
-    static let isSideloaded = Bundle.main.bundleIdentifier != "app.aidoku.Aidoku"
+    #if CANONICAL_BUILD          // true only for App-Store scheme
+        static let canonicalID = "app.aidoku.Aidoku"
+    #else
+        static let canonicalID = Bundle.main.bundleIdentifier ?? ""
+    #endif
+
```
Skittyblock, line 18:
```suggestion
#if CANONICAL_BUILD          // true only for App-Store scheme
    static let canonicalID = "app.aidoku.Aidoku"
#else
    static let canonicalID = Bundle.main.bundleIdentifier ?? ""
#endif

```

iOS/AppDelegate.swift
```diff
@@ -10,8 +10,13 @@ import Nuke
...
 class AppDelegate: UIResponder, UIApplicationDelegate {
-
-    static let isSideloaded = Bundle.main.bundleIdentifier != "app.aidoku.Aidoku"
+    #if CANONICAL_BUILD          // true only for App-Store scheme
+        static let canonicalID = "app.aidoku.Aidoku"
+    #else
+        static let canonicalID = Bundle.main.bundleIdentifier ?? ""
+    #endif
+    
+    static let isSideloaded = Bundle.main.bundleIdentifier != canonicalID
```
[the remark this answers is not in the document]
nschumac: From your point of view it doesn‘t change much, you compile the program with the flag set in the xcconfig. And it would just keep the normal check. I‘ve checked this.
For those who want to compile it themselves and use it they can remove the check with a simple edit of the xcconfig.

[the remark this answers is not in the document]
Skittyblock: oh i see. it's just if you remove the `CANONICAL_BUILD` flag. that's fine then

Skittyblock (approved):
thanks, sorry it took so long to get this in

PR #587, opened 2025-05-21
Add AltStore source badges in README

imkh:
Taking inspiration from repos such as https://github.com/Balackburn/Apollo and https://github.com/lo-cafe/winston, I’ve added badges that make it easy to quickly add the AltStore source on an iOS device.

https://github.com/user-attachments/assets/8e57f9cd-b536-45b7-ac8d-1a6abae844c1

CleanShot 2025-05-21 at 20 51 53

Regarding the ⁠`https://intradeus.github.io/http-protocol-redirector?r` URL: this is necessary because GitHub restricts the use of non-http(s) schemas for security reasons. Using https://github.com/intradeus/http-protocol-redirector should be safe, but I’ll let you decide.

Also, here are a few other options. Let me know if you’d like me to add/change, or if the setup shown in the screenshot above works for you:
1. I can replace the ⁠`Add to AltStore` badge with an `⁠Add to AltStore Classic` if you prefer. I find it a bit too wide but it makes it 100% clear that the source is not compatible with AltStore PAL.
2. The badges are currently set to a height of 60. I can reduce this to 40 or 50 if you prefer them shorter.
3. Should I also add the badges to the https://aidoku.app website? If yes, should I move the images I added in this repo to the website's repo? The [static/images/](https://github.com/Aidoku/Website/tree/main/static/images) directory would make more sense.

CleanShot 2025-05-21 at 20 57 38


Skittyblock:
hi, thank you for the contribution, but to be honest i don't think this is something i really want in the github readme. i'm just personally not a huge fan of badges/graphics of any sort in readmes. this sort functionality does make more sense on the website, and the redirector wouldn't be necessary, but i think there the images wouldn't be necessary since they could be done through css instead.

my apologies for closing this...

imkh:
No worries, I understand. I'd still like to have the badges somewhere for convenience when (re)installing Aidoku. Would you accept the PR if I opened it on the website's repo? Not quite sure what you mean by `images wouldn't be necessary since they could be done through css instead` however. I was thinking of adding the images in the [static/images/](https://github.com/Aidoku/Website/tree/main/static/images) directory (either `png` or `svg` as you prefer.)

Skittyblock:
well i thought that there wouldn't be much reason to use images as opposed to some styled html elements, since it's relatively basic styling (border radius, stroke, text color, and alignment). if the svgs are small they could be considered, but it makes more sense to me to have html elements since they're easily modifiable in the future if necessary

PR #612, opened 2025-06-25
Implement Image Reload - Issue Aidoku#423

### LINKED ISSUE #423: Manually Reload Pages
When the pages are still loading (showing a loading indicator) and users switch to other apps, the images might fail to be decoded and show grey blocks of pixels or have noise after the loading process is completed. Having a feature to manually force the pages to reload would be useful since there is no way to clear the cache of a manga or a chapter.

## Screenshots

IMG_1819

## Environment

- Aidoku version: 0.6.7(1)
- iOS version: 17.4.1
- Device model: iPhone SE (2nd generation)

realdoomsboygaming:
Mistakenly pushed to wrong branch and then hard reset to wrong commit, sorry for the additional PR

Closes #423

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -459,6 +459,8 @@
 "SUPPORT_VIA_KOFI" = "Support via Ko-Fi";
 "UNREAD_CHAPTERS" = "Unread Chapters";
 "RELOAD" = "Reload";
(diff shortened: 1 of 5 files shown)


Skittyblock:
looks like there's some whitespace left on lines; xcode does this by default which is annoying, you can turn on "trim whitespace-only lines" on the editing tab in the settings page

PR #620, opened 2025-06-26
Feat: Add Download Manager for downloaded files

realdoomsboygaming:
This PR introduces a string that will need to be added and localized

DOWNLOAD_MANAGER


Skittyblock (approved):
thank you for the pr! i think this should be good to go now, but i'll do a little more testing before merging

for reference, here are the changes i made:
- moved into separated settings tab section
- added localization for everything
    - “manga” -> “series” (trying to be a more general content reader, not limited to manga)
    - “1 chapters” -> “1 chapter” (there’s a special localization string for this)
- removed confirmation for deleting single chapters
- used static properties for the notification names (e.g. .downloadRemoved instead of NSNotification.Name("downloadRemoved”))
- moved new views into an organized folder
- split view models into separate files
    - put view models under the corresponding view namespace instead of in new classes
- made reverse sort button tinted
- adjusted “in library” label (app’s pink tint not blue, smaller image, less spacing, dynamic font)
- .navigationBarTrailing -> .topBarTrailing
- added remove all downloads (like all of em all of em) option
- formatted chapters titles like the manga page
- removed visible trash button, just using swipe to delete chapters
- used onDelete for the swipe action, not a manual swipe action
- made chapters readable from the download page
- used dynamic text sizes instead of manual
- used foregroundStyle instead of foregroundColor (deprecated)
- AsyncImage -> MangaCoverView (already had a view for this)
- used new (swiftui) manga view instead of mangaviewcontroller
- disabled opening manga page for items not in library
- used NavigationLink for downloaded manga list

PR #650, opened 2025-08-03
chore(github): add issue templates for bugs and features

remytomme:
Adds templates to guide users when creating issues and to automatically apply the `bug` and `enhancement` labels.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: "[BUG]"
+labels: bug
+assignees: Skittyblock
+
+---
+
+### Summary
+
+Summarize the bug encountered concisely.
+
+### Steps to reproduce
+
+Describe how one can reproduce the issue — this is crucial. Please use an ordered list.
+
+### What is the current *bug* behavior?
+
+Describe what actually happens.
+
+### What is the expected *correct* behavior?
+
+Describe what you should see instead.
+
+### Relevant log output and/or screenshots
+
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,108 @@
+name: 🐞 Bug report
```
Skittyblock, line 1: i don't think we need any emojis
```suggestion
name: Bug report
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,108 @@
...
+      placeholder: |
+        Example: "App crashes when opening manga details"
+    validations:
+      required: true
+
+  - type: textarea
+    id: reproduce-steps
+    attributes:
+      label: Steps to reproduce
+      description: Describe how one can reproduce the issue — this is crucial. Please use an ordered list.
```
Skittyblock, line 20:
```suggestion
      description: Describe how one can reproduce the issue.
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: false
```
Skittyblock, line 1: we should still allow blank issues just in case
```suggestion
blank_issues_enabled: true
```

Skittyblock (changes requested):
thank you! this is pretty much good, just a couple things:
- we don't need emojis in the titles
- the only required part of the bug report should be the bug summary. there are cases when steps to reproduce aren't applicable, along with the rest of the form elements.

i might tweak it a bit after it gets merged though since i can't see exactly how it looks yet

Issue #445, opened 2024-05-18
Ability to offset spreads


fancyham:
Ability to offset the pages so that odd pages are on the left (when reading right to left).  On mangadex, this is the “o” key binding.

Would be nice to have a simple gesture like two-finger swipe, and to have a keyboard command,

Skittyblock:
probably goes along with #95

fancyham:
Similar (or maybe same) — my request would be some shortcuts to toggle this on/off because sometimes chapters start with inconsistent pages (i.e. sometimes the left page or sometimes the right page) due to padding or just how it was originally printed.

(See mangadex’s reader for examples of how this might work)

fancyham:
Paperback does this by press and hold on a page, then one of the contextual options is to display that page by itself.

It’s not perfect but at least lets the user display spreads correctly for following pages.

A better option would be to contextual press on a page and select it as the ‘leading’ page., I.e. showing it on the right if I’m reading right-to-left.  Or simply an option to toggle between starting this chapter with a 2-up spread or starting with a 1-up page.

Ari-03:
I think having this would greatly increase aidoku's double page readability.

PR #697, opened 2025-09-21
feat: implement wide image detection and page offset functions in double page layout

### LINKED ISSUE #96: [Enhancement] Double page auto determine wide pages

dyphire:
Closes #96
Partial fix for #445

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -509,6 +509,7 @@
 "CACHE_SIZE_%@" = "Cache size: %@";
 "SUPPORT_VIA_KOFI" = "Support via Ko-Fi";
 "UNREAD_CHAPTERS" = "Unread Chapters";
+"SET_AS_SINGLE_PAGE" = "Set as Single Page";
 "RELOAD" = "Reload";
 "RELOAD_FAILED" = "Reload Failed";
 "RELOAD_FAILED_TEXT" = "Failed to reload the image. Please try again.";
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -469,6 +469,7 @@
 "PAGED" = "Paged";
 "PAGE_LAYOUT" = "Page Layout";
 "DOUBLE_PAGE" = "Double Page";
+"ISOLATE_FIRST_PAGE" = "Always Isolate First Page";
 "MULTI_LANGUAGE" = "Multi-Language";
 
 // Tracking
--- a/iOS/New/Views/Reader/ReaderSettingsView.swift
+++ b/iOS/New/Views/Reader/ReaderSettingsView.swift
@@ -204,6 +204,14 @@ struct ReaderSettingsView: View {
                                 ))
                             )
                         )
+                        SettingView(
+                            setting: .init(
+                                key: "Reader.pagedIsolateFirstPage",
+                                title: NSLocalizedString("ISOLATE_FIRST_PAGE"),
+                                notification: .init("Reader.pagedIsolateFirstPage"),
+                                value: .toggle(.init())
+                            )
+                        )
                     }
                 }
 
--- a/iOS/UI/Reader/Readers/Paged/ReaderPagedViewController.swift
+++ b/iOS/UI/Reader/Readers/Paged/ReaderPagedViewController.swift
@@ -141,6 +141,9 @@ extension ReaderPagedViewController {
             } else {
                 let page = ReaderPageViewController(type: .page)
                 page.pageView?.imageView.addInteraction(UIContextMenuInteraction(delegate: self))
+                page.onAspectRatioUpdated = { [weak self] in
+                    self?.handleAspectRatioUpdate()
+                }
                 pageViewControllers.append(page)
             }
         }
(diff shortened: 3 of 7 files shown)


Shared/Localization/en.lproj/Localizable.strings
```diff
@@ -469,6 +469,7 @@
 "PAGED" = "Paged";
 "PAGE_LAYOUT" = "Page Layout";
 "DOUBLE_PAGE" = "Double Page";
+"ISOLATE_FIRST_PAGE" = "Isolate First Page";
```
Skittyblock, line 472: rather than labelling this as a setting to isolate the first page (which could already be isolated due to the first or second page being wide), i think we should copy mangadex's name for the setting. it seems to behave the same way.
```suggestion
"OFFSET_DOUBLE_SPREADS" = "Offset Double Spreads";
```

Shared/Localization/en.lproj/Localizable.strings
```diff
@@ -507,6 +508,7 @@
 "CACHE_SIZE_%@" = "Cache size: %@";
 "SUPPORT_VIA_KOFI" = "Support via Ko-Fi";
 "UNREAD_CHAPTERS" = "Unread Chapters";
+"AS_ISOLATE_PAGE" = "As Isolate Page";
```
Skittyblock, line 511: i think a more natural sounding name for this would be something like this. isolate is technically the right word but a little awkward used in this scenario
```suggestion
"SET_AS_SINGLE_PAGE" = "Set as Single Page";
```

iOS/New/Views/Reader/ReaderSettingsView.swift
```diff
@@ -204,6 +204,14 @@ struct ReaderSettingsView: View {
                                 ))
                             )
                         )
+                        SettingView(
+                            setting: .init(
+                                key: "Reader.pagedIsolateFirstPage",
+                                title: NSLocalizedString("ISOLATE_FIRST_PAGE"),
+                                notification: .init("Reader.pagedIsolateFirstPage"),
+                                value: .toggle(.init())
+                            )
```
Skittyblock, line 213: this setting should also be added to the `Settings.readerSettings` static variable to show in the settings tab

iOS/UI/Reader/Readers/Paged/ReaderPagedViewController.swift
```diff
@@ -574,6 +688,15 @@ extension ReaderPagedViewController: UIContextMenuInteractionDelegate {
...
             }
 
+            let isolatePageAction = UIAction(
+                title: NSLocalizedString("AS_ISOLATE_PAGE", comment: ""),
+                image: UIImage(systemName: "rectangle.portrait")
+            ) { _ in
+                Task { @MainActor in
+                    self.isolateCurrentPage(for: pageView)
+                }
+            }
```
Skittyblock, line 698: could we only add this if we're using the double page layout currently? would also be best to not show it if we've already isolated the selected page view (in the case the option does nothing)

iOS/UI/Reader/Readers/Paged/ReaderPagedViewModel.swift
Skittyblock, line 1: i pushed a commit to remove these settings variables since they're no longer used, so you should pull main and remove this to fix the merge conflict

Skittyblock (changes requested):
thank you very much for getting this started!! i've wanted to add it for a while but never gotten around to it

Shared/Localization/en.lproj/Localizable.strings
dyphire: Suwatte named this feature Always Isolate First Panel, while Paperback named this feature Always Separate First Page. 
I personally think this naming method is more appropriate because its application method is to offset from the first page.

iOS/UI/Reader/Readers/Paged/ReaderPagedViewModel.swift
Skittyblock: hmm, it seems to work fine on my end. what version are you using? do you have a crash log?

Shared/Localization/en.lproj/Localizable.strings
dyphire: > consider the following situation: if we have pages like [skinny, wide, skinny, skinny, ...], then the first page will always be isolated regardless of the current setting's state. the page layout would be [single, single, double, ...]. however, if we had an offset setting then it would isolate the first page of the first double layout, making it [single, single, single, double, ...] (and would also isolate the first page in the same way, achieving the same thing). in order to achieve the same behavior with the current setting the user would need to manually set one of the pages as a single page.

Yes, I have taken this scenario and approach into account. But there's another problem: intelligently detecting the first double-page layout and then setting up a single page will detect from beginning to end and won't trigger if you encounter the single page display all of images is wide.

Taking into account the above I added a third commit to allow for more flexibility in adjusting page offsets.

> anyways, you can probably leave the setting for now and i might try and replace it with the offset setting in the future. you should add in the "always" text to make it clearer then:

Done.

iOS/UI/Reader/Readers/Paged/ReaderPagedViewModel.swift
Skittyblock: thanks for spotting that, i just pushed a fix to main

dyphire: Now it works fine.

Skittyblock (approved):

PR #711, opened 2025-09-30
feat: add image splitting for wide images in reader

### LINKED ISSUE #698: The single-page layout provides the function of splitting wide drawings into double pages
### Feature Description

Mark, plan to implement.

### Acknowledgements

dyphire:
Close #698

Page index sometimes has problems, need help.


Skittyblock:
i pushed a fix for the previous images not splitting (it's not perfect but it works enough). it does seem that there are some times where it gets stuck on a section of pages and breaks moving off of them, but i can't figure out a consistent way to trigger that -- do you know? i think it's similar to the bug where chapter switching doesn't work if you're scrolling too fast

dyphire:
> i pushed a fix for the previous images not splitting (it's not perfect but it works enough). 

I've tried implementing it before, but something went wrong so I gave up. Your implementation LGTM.

> it does seem that there are some times where it gets stuck on a section of pages and breaks moving off of them, but i can't figure out a consistent way to trigger that -- do you know? i think it's similar to the bug where chapter switching doesn't work if you're scrolling too fast

I also couldn't reproduce it stably, the only thing I was sure of is that when this happens, switch the option of "split width image" will return to normal once. So I suspect there is a hidden problem with the page index.

PR #748, opened 2025-10-18
feat: implement custom Home-like layout for Komga listings

dyphire:
Before:
image image
After:
image image


dyphire:
Rebase

dyphire:
> not a huge fan of the incorrect skeleton loading view but it's probably going to far to correct it

I also don't like the incorrect skeleton loading view, but it's a bit expensive to correct it, especially when the library only has one and still need to use this skeleton.

Skittyblock:
oh my bad, thanks for the lint fix

PR #758, opened 2025-10-21
feat: add close swipe gesture for reader view

### LINKED ISSUE #709: Swipe to close reader view.
### Feature Description

Just like the photo's app. Swipe down to close the reader view when in any horizontal reading mode, or right to close webtoon mode. This makes navigating the app with one hand a lot easier.

### Acknowledgements

dyphire:
Close #709

Demo:

https://github.com/user-attachments/assets/b5d541af-7c29-4fe3-be00-548d65e1f2e8

https://github.com/user-attachments/assets/5c94dc3e-3e6a-4d2d-a928-c31eaf628772

### DIFF
--- a/iOS/UI/Reader/ReaderViewController.swift
+++ b/iOS/UI/Reader/ReaderViewController.swift
@@ -70,6 +70,38 @@ class ReaderViewController: BaseObservingViewController {
         return tap
     }()
 
+    private lazy var closeSwipeGestures: [UISwipeGestureRecognizer] = {
+        let directions: [UISwipeGestureRecognizer.Direction] = [.left, .right, .down]
+        return directions.map { direction in
+            let swipe = UISwipeGestureRecognizer(target: self, action: #selector(handleCloseSwipe(_:)))
+            swipe.direction = direction
+            return swipe
+        }
+    }()
+
+    private lazy var closePanGesture: UIPanGestureRecognizer = {
+        let pan = UIPanGestureRecognizer(target: self, action: #selector(handleClosePan(_:)))
+        return pan
+    }()
+
+    private var initialCenter: CGPoint = .zero
+    private var isDismissing = false
+
+    private func updateSwipeGestures(for readingMode: ReadingMode) {
+        switch readingMode {
+        case .ltr, .rtl:
+            // Horizontal paged modes: only down swipe enabled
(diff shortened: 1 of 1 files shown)


Skittyblock (changes requested):
while this is interesting and i can see it being useful, it isn't very apple-like, so i'm not sure if i could really approve adding it. 

i think the closest example of what we could do is a specific small list view in the news app that expands into the article view, but the same type of behavior is used across nearly all of the apple apps (in ios 26 at least). i'm not sure if there's an easy built-in way to do it with swiftui/uikit (there must be), but in order to have the swipe to dismiss on the reader view, it should expand entirely from the chapter cell, and swiping should move the entire sheet view instead of just the page.

if this type of behavior is only available ios 26+ that would probably be fine, then we could fall back to this pr's implementation for older ios versions. in that case, the velocity may need a little adjustment; for example, when swiping down, holding, then swiping up quickly, the view often still closes. this seems a little unnatural to me, and i think it probably makes sense to *not* dismiss if there's any upwards velocity at all.

dyphire:
> while this is interesting and i can see it being useful, it isn't very apple-like, so i'm not sure if i could really approve adding it.
> 
> i think the closest example of what we could do is a specific small list view in the news app that expands into the article view, but the same type of behavior is used across nearly all of the apple apps (in ios 26 at least). i'm not sure if there's an easy built-in way to do it with swiftui/uikit (there must be), but in order to have the swipe to dismiss on the reader view, it should expand entirely from the chapter cell, and swiping should move the entire sheet view instead of just the page.

I think implementing the idea you mentioned may require a lot of UI refactoring, and maybe we can provide the current solution before trying to improve it.

> if this type of behavior is only available ios 26+ that would probably be fine, then we could fall back to this pr's implementation for older ios versions. in that case, the velocity may need a little adjustment; for example, when swiping down, holding, then swiping up quickly, the view often still closes. this seems a little unnatural to me, and i think it probably makes sense to _not_ dismiss if there's any upwards velocity at all.

Implement the suggestions on velocity.

Skittyblock:
> maybe we can provide the current solution before trying to improve it

nah the reason to add it would be just if it were only on the non-mainline ios versions. i don't care as much about the older ios versions, as long as the app generally works, since it's a much smaller fraction of users. we can wait until i have time to look into the full view dismiss gesture.

PR #761, opened 2025-10-22
fix: correct clear souce cache i18n key

aphronyx:
## Changes

- Fixed an issue where the `Clear Source Cache` button in source setting was not properly localised
  | Before | After |
  | :-: | :-: |
  | before-setting | after-setting |

## Notes


Skittyblock (approved):
oops, thanks

PR #762, opened 2025-10-22
fix: use inline title for `AddSourceView`

aphronyx:
## Premise

The `navigationBarTitleDisplayMode` for `AddSourceView` is not set to `inline`, which is inconsistent with the other sheet views (e.g. `MigrateSourcesView`).

## Changes

- Fixed an issue where the navigation title of `AddSourceView` was not inline
  | Before | After |
  | :-: | :-: |
  | add_source-before | add_source-after |

## Notes

### DIFF
--- a/iOS/New/Views/Browse/AddSourceView.swift
+++ b/iOS/New/Views/Browse/AddSourceView.swift
@@ -184,6 +184,7 @@ struct AddSourceView: View {
                 }
             }
             .navigationTitle(NSLocalizedString("ADD_SOURCE"))
+            .navigationBarTitleDisplayMode(.inline)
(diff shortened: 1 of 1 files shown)


Skittyblock:
while it is true that inline titles for sheets are more conventional and that's what we *should* do.... i really don't like the additional space between the title and the import source button. i think i need to come up with a different design for it or something that looks better before merging.

aphronyx:
> i think i need to come up with a different design for it or something that looks better before merging.

oh sorry, I thought it was a mistake during the migration to the 0.7 new views. there are actually more sheet views that use large titles in aidoku. i should probably close this pull request and open an issue instead.

PR #767, opened 2025-10-25
Add swipe action to delete manga update

kaan-w:
### Demo
https://github.com/user-attachments/assets/8aaab9b8-06bb-4065-9714-e30caa48646e

### DIFF
--- a/iOS/UI/Updates/MangaUpdatesView.swift
+++ b/iOS/UI/Updates/MangaUpdatesView.swift
@@ -19,9 +19,7 @@ struct MangaUpdatesView: View {
     }
     struct UpdateInfo: Identifiable, Hashable {
         let id: String
-        let sourceId: String
-        let chapterId: String
-        let mangaId: String
+        let chapterIdentifier: ChapterIdentifier
         let date: Date
         let manga: AidokuRunner.Manga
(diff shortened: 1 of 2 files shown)


iOS/UI/Updates/MangaUpdatesView.swift
```diff
@@ -92,17 +92,23 @@ struct MangaUpdatesView: View {
...
                             MangaUpdateItemView(updates: updates)
                         }
                         .offsetListSeparator()
                         .id(item.mangaKey)
+                        .swipeActions(edge: .trailing) {
+                            Button(NSLocalizedString("DELETE")) {
+                                removeUpdate(update: update, day: entry.day)
+                            }
+                            .tint(.red)
+                        }
```
Skittyblock: can we use `onDelete` instead of `swipeActions` for the system styling? if not, we should copy the style in `BackupsView` and use a destructive button with a `Label` with the trash icon.

wvy777:
- The CoreData function now uses an array of tuples containing `sourceId`, `mangaId`, and `chapterId` to delete the objects.

### New Button Demo
https://github.com/user-attachments/assets/d51022f0-bb21-4168-82d9-767e86366240

PR #778, opened 2025-11-09
Prevent unlocking when biometric authentication is disabled

### LINKED ISSUE #358: Pin/Passcode feature for Library Locking for those without faceID/fingerprint
Not only for those people who lack the FaceID/Fingerprint features on their devices, but also for use in environment(s) and/or conditions where the features don't work.

### LINKED ISSUE #777: User can bypass app locking by disabling biometric authentication
### Summary

Similar but distinct to #358, a user can disable library and history locking by disabling biometric authentication. I presume this is due to Aidoku treating biometrics unavailable as unexpected, since [the fallback unlocks the app regardless](https://github.com/Aidoku/Aidoku/blob/68a19523cc4a4518cde1f70351847aa2d6bafcfa/iOS/UI/Library/LibraryViewController.swift#L700-L704).

### Steps to reproduce

1. Open the app
2. Select a library category (or just all) or the history tab
3. Continuously fail to authenticate
4. Disable biometric authentication
5. Try to authenticate and see the app unlock

When biometric authentication is disabled, the app unlocks.

The app should remain locked, which the user can resolve by re-authenticating the device (say, from the lock screen).

### Screenshots

https://github.com/user-attachments/assets/432465b7-d856-43f4-b21c-9f0231253bd9

### Device

iPhone 12

### Operating System

iOS

### OS Version

26.1

### App Version

0.7.2 (build 2)

### Acknowledgements

kyleerhabor:
Closes #777.
Closes #358.

While we're here, it may be worth it to address #358, which can be resolved by changing `LAPolicy.defaultPolicy`'s value to `LAPolicy.deviceOwnerAuthentication`.

### DIFF
--- a/Shared/Extensions/LAPolicy.swift
+++ b/Shared/Extensions/LAPolicy.swift
@@ -8,5 +8,5 @@
 import LocalAuthentication
 
 extension LAPolicy {
-    static let defaultPolicy = Self.deviceOwnerAuthenticationWithBiometrics
+    static let defaultPolicy = Self.deviceOwnerAuthentication
 }


kyleerhabor:
@Skittyblock I've made the change. Mind adding the issue to this pull request like #777 is?

Issue #817, opened 2025-11-20
Opening global search filters view breaks keyboard


Skittyblock:
it dismisses and won't re-present unless you exit the tab and re-enter

PR #840, opened 2025-11-26
Add manual refocus for keyboard when close filter

dangobums:
Fixing issue https://github.com/Aidoku/Aidoku/issues/817


Skittyblock:
it looks like this will focus the keyboard even if it wasn't focused before opening the sheet (e.g. if you search something first, then open the filters sheet to adjust filters, it then reopens the keyboard on close)

is it possible to keep track of this?

dangobums:
i think we can use searchCommitToggle in the SearchContentView to keep track or we just release the focus of the keyboard and let the user control the focus themself

PR #846, opened 2025-12-02
fix: XML requires that quotes and apostrophes be escaped

NightQuest:
per the [spec](https://www.w3.org/TR/REC-xml/#syntax):
`To allow attribute values to contain both single and double quotes, the apostrophe or single-quote character (') may be represented as " &apos; ", and the double-quote character (") as " &quot; ".`


Skittyblock (approved):
the reason i didn't have this was because i thought most xml parsers were smart enough to understand that `"` and `'` appearing between tags shouldn't be confused with tag attributes(?), so i'm surprised it makes a difference. i thought the escaping was only necessary within attributes themselves.

i ran the tests on main to double check and it doesn't break the current parsing though, so it's fine to merge. thanks.

PR #855, opened 2025-12-16
feat: expose webview localstorage to sources

### LINKED ISSUE #854: Expose WebView LocalStorage to Sources
## Description

Allow sources to access `localStorage` data from the WebView component.

## Use Case

Many web authentication flows store tokens in `localStorage` instead of (or alongside) cookies. Currently, there's no way for sources to extract this data during WebView interactions (e.g., login).

## Proposed Behavior

Extract specified `localStorage` entries after relevant WebView events and expose them to sources via the existing `defaults` system, similar to cookie handling.

## Impact

- Adds new functionality (SemVer MINOR)
- No breaking changes

remytomme:
closes #854

### DIFF
--- a/iOS/New/Views/Common/Settings/WebView.swift
+++ b/iOS/New/Views/Common/Settings/WebView.swift
@@ -10,18 +10,21 @@ import WebKit
 
 struct WebView: UIViewRepresentable {
     let url: URL
+    var sourceId: String?
 
     @Binding var cookies: [String: String]
     @Binding var reloadToggle: Bool
 
-    private let webView = WKWebView()
+    let webView = WKWebView()
 
     init(
         _ url: URL,
+        sourceId: String? = nil,
         cookies: Binding<[String: String]> = .constant([:]),
         reloadToggle: Binding<Bool> = .constant(false)
     ) {
         self.url = url
+        self.sourceId = sourceId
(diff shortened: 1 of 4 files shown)


remytomme:
**Test Details:**

* **Device:** iPhone 16 Pro Max
* **OS:** iOS 18.7.2

**Status:** Functionality verified, works as intended.
**Ready for review.**

iOS/New/Views/Common/Settings/WebView.swift
```diff
@@ -57,16 +62,32 @@ struct WebView: UIViewRepresentable {
         }
 
         func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
-            Task {
+            Task { @MainActor in
+                if let sourceId = parent.sourceId {
+                    let localStorage = await webView.getLocalStorage()
+                    for (key, value) in localStorage {
+                        UserDefaults.standard.set(value, forKey: "\(sourceId).\(key)")
```
Skittyblock: instead of passing in the source id and assigning to the user defaults here, it would be more ideal to do the same thing as cookies and pass in a binding for the local storage keys, then assign them in the same place as the cookies are. it would be better if we could namespace them with the login settings key like cookies.

PR #858, opened 2025-12-17
feat: support extracting localstorage in webview login

### LINKED ISSUE #854: Expose WebView LocalStorage to Sources

remytomme:
Extracts keys defined in `localStorageKeys` during WebView login and saves them to UserDefaults using a namespaced key.

closes #854

### DIFF
--- a/Aidoku.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Aidoku.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -7,7 +7,7 @@
       "location" : "https://github.com/Aidoku/AidokuRunner",
       "state" : {
         "branch" : "main",
-        "revision" : "1b9f9be34ca1463aabae053f28e574d3bbef03b3"
+        "revision" : "b155fb8de83cb3671e6bb4d59cd6db63e2290e7b"
       }
     },
     {


iOS/New/Views/Common/Settings/SettingView.swift
```diff
@@ -762,6 +764,12 @@ extension SettingView {
                 SettingsStore.shared.remove(key: key + Self.passwordKeySuffix)
                 SettingsStore.shared.remove(key: key + Self.cookieKeysKeySuffix)
                 SettingsStore.shared.remove(key: key + Self.cookieValuesKeySuffix)
+                // remove local storage
+                if let sourceId = source?.id, let localStorageKeys = value.localStorageKeys {
+                    for key in localStorageKeys {
+                        SettingsStore.shared.remove(key: "\(sourceId)\(Self.localStoragePrefix)\(key)")
```
Skittyblock: we should namespace it with the `key` prefix like the other ones, since technically a source could use multiple different web login settings and we wouldn't want the local storage keys to possible conflict.

PR #861, opened 2025-12-17
fix setting button not going to root when tapped

### LINKED ISSUE #832: Tapping Settings Tab Not Going to First Page
### Summary

https://github.com/user-attachments/assets/022d6fb5-6814-4211-89a9-e1977eca8497

### Steps to reproduce

_No response_

### Device and version

_No response_

686udjie:
closes #832

### DIFF
--- a/iOS/UI/Common/TabBarController.swift
+++ b/iOS/UI/Common/TabBarController.swift
@@ -234,7 +234,7 @@ extension TabBarController {
 
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
-        
+
         if #unavailable(iOS 26.0) {
             let height: CGFloat = 48
             let padding: CGFloat = 16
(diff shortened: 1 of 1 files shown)


Skittyblock (changes requested):
the reason i switched to the swiftui navigation was to fix the backups view toolbar button zoom transitions. however, i accidentally left it as a `NavigationView` instead of `PlatformNavigationStack` while testing so it was broken anyways.

is there any way we could fix the zoom transitions, but still keep the tab bar tap to pop to root functionality?

iOS/UI/Common/TabBarController.swift
```diff
@@ -244,6 +233,8 @@ extension TabBarController {
     }
 
     override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
```
Skittyblock: not much point calling this since the default impl does nothing https://developer.apple.com/documentation/uikit/uiviewcontroller/viewdidlayoutsubviews()

PR #876, opened 2026-01-01
persistent scroll position for webtoon reader

### LINKED ISSUE #676: Webtoon Mode Stores Wrong Page
### Summary

When the reading mode is set to webtoon, reader stores the wrong page.

### Steps to reproduce

1. Set reading mode to webtoon
2. Open a chapter
3. Close reader view

_No response_

_No response_

### Screenshots

https://github.com/user-attachments/assets/c831787d-e561-4808-aee1-d085fe319a16

### Device

iPhone 16 Pro Simulator

### Operating System

iOS

### OS Version

26.0 beta 6 (23A5326a)

### App Version

c0a14e0

### Acknowledgements

686udjie:
partially fixes  #676 

adds scroll position saving and restoring for webtoon reader chapters

p.s. this sometimes doesnt work for whatever reason, but i think its better for it to be working for most situations then none at all. 

how it works:
https://litter.catbox.moe/yllbl040vp0o18q9.mov

When reading a chapter that’s already marked as read, I don’t save the offset, so reopening it jumps to the last page again.
If you mark it unread, it starts saving progress.

Should I keep it like this, or should it remember whether the chapter was read?

### DIFF
--- a/iOS/UI/Reader/Readers/Webtoon/ReaderWebtoonViewController.swift
+++ b/iOS/UI/Reader/Readers/Webtoon/ReaderWebtoonViewController.swift
@@ -191,26 +191,16 @@ extension ReaderWebtoonViewController {
         }
 
         // update slider position
-        if let layout = self.collectionNode.collectionViewLayout as? VerticalContentOffsetPreservingLayout,
-           let currentPages = pages[safe: chapterIndex] {
-            var offset: CGFloat = 0
-            for idx in 0..<chapterIndex {
-                offset += layout.getHeightFor(section: idx)
+        if let currentPages = pages[safe: chapterIndex] {
+            let imagePages = currentPages.filter { $0.type == .imagePage }
+            let totalPages = imagePages.count
+            let page = getCurrentPage()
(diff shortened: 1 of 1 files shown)


Skittyblock (changes requested):
a few things:
- i think storing the exact scroll offset *is* the ideal solution, if it can be made to work correctly
- however, it seems like the main issue is that this adds a noticeable delay when loading a chapter in the webtoon reader. is this avoidable somehow?
- using userdefaults to store this kinda thing isn't a great idea, especially since you're not including the source key so there could be conflicts there. i added a coredata attribute, `scrollPosition`, for history objects which can be used instead
  - the new attribute will be included in the next update to avoid schema conflicts (since i'm already updating the schema for 0.8), but i don't think this pr will make it in before i push the update. it can still come later though
- position should be cleared if the chapter is read in a different reading mode, etc. and the webtoon reader should always fall back to loading based on the starting page if there's no scroll offset. it also should start at the beginning (not load any start position) if the chapter is already completed

Issue #695, opened 2025-09-19
Functionality for Mass-Import and Multiple Selection of Series in a Source


templeofshadow:
### Feature Description

I realize that this may be niche, but here's my use case:

I have two large Komga libraries, and I want to import all of the series from both into my Aidoku library. 

As of right now, as far as I can tell, the only way to do this is to go through each series individually and manually add it to the Aidoku library, along with configuring the category. This would be extremely inconvenient and time-consuming to do, as these libraries combined have upwards of 4.4k series (about 4.1k of them are single-book doujinshi, so it's not actually that much data).

**What I'd like to see is a "select multiple series" function in the source browse page, and perhaps if the source is self-hosted or has a limited number of series, a "select all" button, and functions for adding all selected series into the Aidoku library.** This "select multiple" functionality would work per-page, so if you have multiple Komga libraries like me, the function would only work on series in the library whose page you are working under (as in- the Komga source separates libraries into the "Home", "Library 1"... etc. tabs; when I say "page", I mean those library tabs). It would probably also be good to add this functionality to the source search page.

There appears to already be similar functionality with the Aidoku library page itself, so it seems possible to do the same in the source browse page. 

It might also be interesting to add some kind of "mirror source library" functionality, so that if you use solely Komga/Kavita/etc. as your source, and do all manga sourcing on the server side, you can simply mirror the source in your Aidoku library. I believe Komga has categories and such, and Kavita has Collections (which aren't the same thing but similar enough to serve the same function most of the time) so you could even mirror library categories. 

This would be quite useful for library management, because frankly, managing a library in a mobile app is inconvenient on account of the limited screen real-estate and input methods. If the source has multiple libraries (like my setup), you could choose to merge the libraries or maybe create some kind of super-category for each library, which then syncs the categories of the source as sub-categories under each library category. This sounds complicated though, so take it with a hefty grain of salt.

In any case, at the very least, there should probably be a faster way of adding a series to the library. Maybe a function using a double-tap (like how liking images works in some social media)? The tap-hold is quite slow for adding many series at a given time, although it could work as a context menu (e.g. where you would would put the "select" button for the possible select multiple feature).

In any case, thanks for your time. Apart from this issue (which makes most of my library functionally unusable until it is fixed), I absolutely love the UI and feature set in Aidoku, and would really like to use it as my primary reader app for my library. Keep up the good work.

I apologize if I missed a duplicate issue.

### Acknowledgements

[...]

PR #891, opened 2026-01-12
feat(komga, kavita): add auto track function

dyphire:
Komga and Kavita have complete book management functions themselves, and there is no mandatory need to add bookshelves to use. Let us to provide a better and native experience. 
At the same time, when using shared services, users may not want to enable the automatic tracking of bookshelves, which is also optimized.

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -234,6 +234,9 @@
 "DISPLAY_VOLUME" = "Volume";
 "USE_CHAPTERS" = "Use Chapters";
 "USE_CHAPTERS_TEXT" = "Treat books as chapters instead of volumes";
+"AUTO_ENHANCED_TRACKER" = "Enable Enhanced Tracker";
+"AUTO_ENHANCED_TRACKER_KOMGA_TEXT" = "Automatically track reading progress with enhanced Komga tracker";
+"AUTO_ENHANCED_TRACKER_KAVITA_TEXT" = "Automatically track reading progress with enhanced Kavita tracker";
 "%@_BOOKS" = "%@ Books";
 "1_BOOK" = "1 Book";
 "ON_DECK" = "On Deck";
--- a/Shared/Localization/zh-Hans.lproj/Localizable.strings
+++ b/Shared/Localization/zh-Hans.lproj/Localizable.strings
@@ -241,7 +241,7 @@
 "MULTI_LANGUAGE" = "多语言";
 "TRACKERS" = "追踪器";
 "UPDATE_AFTER_READING" = "阅读后自动更新进度";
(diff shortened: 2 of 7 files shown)


Skittyblock (commented):
i'm not sure if this is something i actually want in the app;
- having the tracker only link automatically when adding to library encourages users to do this (follow the standard/intended usage of the app)
- tracking history for items not in the library reduces app performance (e.g. having to load history in history tab that isn't cached)
- with this pr, tracking would be added on open (rather than e.g. adding the source) which is somewhat undesirable since it's not an action users would expect to have that sort of result (if that makes sense)
- ...and manually removing the tracker creates a default value that can't be removed except in a full setting restore, which isn't great practice and could cause confusion in the future (e.g. if the source is removed and re-added later on)

dyphire:
> i'm not sure if this is something i actually want in the app;
> 
> having the tracker only link automatically when adding to library encourages users to do this (follow the standard/intended usage of the app)
> 
> tracking history for items not in the library reduces app performance (e.g. having to load history in history tab that isn't cached)
> 
>  with this pr, tracking would be added on open (rather than e.g. adding the source) which is somewhat undesirable since it's not an action users would expect to have that sort of result (if that makes sense)

Komga and Kavita are special as sources, as I mentioned earlier, they have full book management and tracking capabilities, and do not depend on the app's bookshelf to use on their own, especially when used cross-platform (win/android/ios).

In this case, the restriction of having to add to the bookshelf for automatic tracking can lead to some bad experiences, such as users who have a large number of one-shot books to read, and now they have to manually add each book to the shelf or manually add tracking to synchronize the reading history to the server.

Also, if the user is using a shared service then them need bookshelf to keep history, but it is also inconvenient for them when the books be added to the bookshelf will automatically track.

I believe there are a lot of Komga and Kavita users who are used to reading directly in the source instead of adding to the bookshelf.

>  ...and manually removing the tracker creates a default value that can't be removed except in a full setting restore, which isn't great practice and could cause confusion in the future (e.g. if the source is removed and re-added later on)

This can indeed become a problem.

Skittyblock:
> I believe there are a lot of Komga and Kavita users who are used to reading directly in the source instead of adding to the bookshelf.

hence why the limitation there is essentially to discourage that, given that the app is not optimized for that in other regards. my main issue with it is that with the current app's tracking system, linking a tracker to an item creates an actual database object; if something is removed from the source at any point (and isn't in the library), then the tracker db object will just end up floating with no connection. however, if the item had been added to the library first, even if it's removed from the source, it must also be removed from the library at some point and thus will automatically unlink and remove the tracker db object. so even though it is possible to link trackers to items not in the library, the ux discourages it.

i think having a large number of one-shots is definitely a valid argument towards the usefulness of something like this, but i still don't think it's enough to convince me. i don't see tapping one button (the bookmark) as being a large inconvenience, as you would only need to sync history if you were planning on reading it in the first place.

if #695 is implemented for komga, it would satisfy this, but that's definitely much more complex of a feature.

PR #904, opened 2026-02-08
feat: add paged text reader

rbertus2000:
Adds a new paginated text reader for text-based sources with a Kindle-like reading experience featuring horizontal page flipping.

Features:
- Horizontal page flipping with smooth transitions
- Two-page spread layout on iPad landscape
- Configurable font family (all system fonts available)
- Configurable font size, line spacing, and horizontal padding
- Reading progress persistence across sessions
- Safe area support for notch/Dynamic Island
- Toggle between new paged mode and original scroll mode

Settings:
- Reader Style: Paged (Kindle) / Scroll
- Font: System fonts picker
- Font Size: 12-32pt
- Line Spacing: 0-24pt
- Horizontal Padding: 8-48pt

Text always reads left-to-right regardless of manga RTL settings.
Settings apply to both paged and scroll text readers.

New files:
- TextPaginator.swift: Core pagination engine using NSLayoutManager
- ReaderPagedTextViewController.swift: UIPageViewController-based reader

### DIFF
--- a/Shared/Localization/de.lproj/Localizable.strings
+++ b/Shared/Localization/de.lproj/Localizable.strings
@@ -333,12 +333,6 @@
 "TEXT_FONT_SIZE" = "Schriftgröße";
 "TEXT_LINE_SPACING" = "Zeilenabstand";
 "TEXT_HORIZONTAL_PADDING" = "Seitenrand";
-"NEXT_CHAPTER" = "Nächstes Kapitel";
-"PREVIOUS_CHAPTER" = "Vorheriges Kapitel";
-"SWIPE_TO_CONTINUE" = "Wischen zum Weiterlesen";
-"SWIPE_TO_GO_BACK" = "Wischen zum Zurückgehen";
-"END_OF_MANGA" = "Du hast das Ende erreicht";
-"START_OF_MANGA" = "Das ist das erste Kapitel";
 
 // Manga Updates
 "MANGA_UPDATES" = "Updates";
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -537,12 +537,6 @@
 "TEXT_FONT_SIZE" = "Font Size";
 "TEXT_LINE_SPACING" = "Line Spacing";
 "TEXT_HORIZONTAL_PADDING" = "Horizontal Padding";
-"NEXT_CHAPTER" = "Next Chapter";
-"PREVIOUS_CHAPTER" = "Previous Chapter";
-"SWIPE_TO_CONTINUE" = "Swipe to continue";
-"SWIPE_TO_GO_BACK" = "Swipe to go back";
-"END_OF_MANGA" = "You've reached the end";
-"START_OF_MANGA" = "This is the first chapter";
 
 // Backups
 "BACKUPS" = "Backups";


iOS/UI/Reader/Readers/Text/TextPaginator.swift
```diff
@@ -0,0 +1,358 @@
...
+        // Calculate content area
+        let contentArea = contentSize(for: pageSize)
+        
+        // Paginate the attributed string
+        return paginateAttributedString(attributedString, contentSize: contentArea, originalMarkdown: markdown)
+    }
+    
+    /// Convert markdown to NSAttributedString with proper styling
+    private func markdownToAttributedString(_ markdown: String) -> NSAttributedString {
+        // Simple markdown conversion - for complex markdown, we'd use a proper parser
```
Skittyblock: not sure if i'm a huge fan of this since it will lose a lot of the markdown capabilities sources may expect, but i guess it's fine if it's just a user-selectable option

maybe this would be better? https://developer.apple.com/documentation/foundation/instantiating-attributed-strings-with-markdown-syntax#overview

iOS/UI/Reader/ReaderViewController.swift
```diff
@@ -730,22 +750,35 @@ extension ReaderViewController: ReaderHoldingDelegate {
...
             showLoadFailAlert()
         } else if pages.count == 1 && pages[0].isTextPage {
-            // single text page, should switch to text reader
-            if !(reader is ReaderTextViewController) {
+            // single text page, should switch to text reader (paginated)
+            if !(reader is ReaderPagedTextViewController) && !(reader is ReaderTextViewController) {
                 setReader(.text)
                 setChapter(chapter)
                 loadCurrentChapter()
+            } else {
```
Skittyblock:
```suggestion
            // single text page, should switch to text reader
            if !(reader is ReaderPagedTextViewController) && !(reader is ReaderTextViewController) {
                setReader(.text)
                setChapter(chapter)
                loadCurrentChapter()
```

iOS/AppDelegate.swift
```diff
@@ -167,6 +167,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 "Reader.pillarboxOrientation": "both",
                 "Reader.orientation": "device",
 
+                // Text Reader defaults
+                "Reader.textReaderStyle": "paged",
+                "Reader.textFontFamily": "Georgia",
```
Skittyblock: not sure if georgia should be the default, i think the system font is better for now

iOS/UI/Reader/Readers/Text/ReaderPagedTextViewController.swift
```diff
@@ -0,0 +1,911 @@
...
+//  Created by Minirob on 2/2/26.
+//
+//  Kindle-style paginated text reader with horizontal page flipping.
+//  Supports single page and two-page spread layouts.
+//
+
+import AidokuRunner
+import UIKit
+import SwiftUI
+import ZIPFoundation
```
Skittyblock: should organize imports alphabetically
```suggestion
import AidokuRunner
import SwiftUI
import UIKit
import ZIPFoundation
```

rbertus2000:
I tried to address all your points. The read position was a good catch. It would change as soon as you change font size, but I solved it with an offset to the saved page. I hope it is good to go now.

rbertus2000:
Hey @Skittyblock, just a friendly ping – any chance you could take another look when you get a moment? All review comments should be addressed. 🙂

rbertus2000:
> it looks like some comments were marked resolved but not addressed(?) i unmarked them resolved, but if you have strong reasoning for not applying them you can reply to the comments explaining
> 
> also you should make sure you have swiftlint installed, or reference the annotations on the files since there are some lint warnings that should be fixed

I resolved all marked issues and there are no lint errors anymore. We now also use virtual pages for scroll view and remember the last reading positon there too.

rbertus2000:
> i tried the changes in the simulator and found these issues, which would need to be fixed before merging:
> 
> * the transitioning between chapters is buggy
>   
>   * if the reader style is scroll, when pressing "continue reading" it switches to the next chapter but keeps the current scroll position at the bottom (and also doesn't split the new chapter into pages)
>   * if paged, when switching to the transition page it just immediately switches to the next chapter in the saved position, rather than the start
>     
>     * ideally, it should work like the image paged reader, where it's a static transition page, and the user can swipe to the next page to continue reading the next chapter
>     * the transition pages should also match the image reader transition pages rather than have a different style
> * single page text chapters aren't marked as read when opened
> * marking chapters as unread then opening them again will still open them at the last page they were at instead of the start
> * markdown isn't processed correctly in the paged reader
>   
>   * there's no styling for headers, which should be possible with attributed strings
>   * newlines are handled incorrectly (regular newlines should be ignored, only lines with two or more trailing spaces or a trailing slash should newline)
> * ideally, we don't want the content to shift up and down when the reader bars are hidden
> * switching between the "paged" and "scroll" type doesn't update unless you close and reopen the reader
> * i'm not entirely sure what the horizontal padding setting does(?) it shifts around my pages but i don't see any visual difference
> 
> i think there are a few more issues but these were the main ones i noticed.

Will look into it.

rbertus2000:
> i tried the changes in the simulator and found these issues, which would need to be fixed before merging:
> 
> * the transitioning between chapters is buggy
>   
>   * if the reader style is scroll, when pressing "continue reading" it switches to the next chapter but keeps the current scroll position at the bottom (and also doesn't split the new chapter into pages)
>   * if paged, when switching to the transition page it just immediately switches to the next chapter in the saved position, rather than the start
>     
>     * ideally, it should work like the image paged reader, where it's a static transition page, and the user can swipe to the next page to continue reading the next chapter
>     * the transition pages should also match the image reader transition pages rather than have a different style
> * single page text chapters aren't marked as read when opened
> * marking chapters as unread then opening them again will still open them at the last page they were at instead of the start
> * markdown isn't processed correctly in the paged reader
>   
>   * there's no styling for headers, which should be possible with attributed strings
>   * newlines are handled incorrectly (regular newlines should be ignored, only lines with two or more trailing spaces or a trailing slash should newline)
> * ideally, we don't want the content to shift up and down when the reader bars are hidden
> * switching between the "paged" and "scroll" type doesn't update unless you close and reopen the reader
> * i'm not entirely sure what the horizontal padding setting does(?) it shifts around my pages but i don't see any visual difference
> 
> i think there are a few more issues but these were the main ones i noticed.

I fixed all mentioned points. Chapter transition is now like Manga/Webtoon Reader. Horizontal padding is now working as intended and you can switch between paged and scroll mode without reopening the reader.

rbertus2000:
Try again, it should be fixed.

rbertus2000:
> formatting is much nicer now, lists and blockquotes look good.
> 
> * still shifting up/down in paged text reader when toggling bars.
> * chapter transitions from a text reader to a non-text reader are broken with the paged style: it seems to switch the chapter in the logic but the pages don't update and everything blacks out. you could disable the transitions in this case, like the scroll style.
> * the "horizontal padding" setting doesn't seem to have any effect on the paged style.

I hope I finally found a solution for the shifting problem. Chapter transitions and horizontal padding are fixed too.

Skittyblock:
thank you for all your changes, this seems good enough functionality-wise. i'm going to take a closer look at the code sometime before the next update and make any other changes myself before merge

PR #906, opened 2026-02-11
Fix/scanlator affinity

kunihir0:
when reading a source that has multiple chapter sources, it was annoying that the default behavior was to choose the first in the list of the next chapter. This change enhances the reader to pick the same chapter source that the reader is using for example,  if user pick scanlator 4, and there is 1,2,3,5 instead of the default behavior being to choose the first available source. This change will attempt to pick the same scanlator 4 chapter source instead of picking scanlator 1 if not then go back to default choose first available.

I was also thinking of maybe adding a drop down menu to the reader settings for the default behavior of choosing the scanlator persistence or not but I dont think that is needed... as the current implementation just reverts back to the default select first available source...

### DIFF
--- a/iOS/UI/Reader/ReaderViewController.swift
+++ b/iOS/UI/Reader/ReaderViewController.swift
@@ -595,6 +595,11 @@ extension ReaderViewController: ReaderHoldingDelegate {
             && (!(a.chapterNumber == nil && a.volumeNumber == nil) || a.title == b.title)
     }
 
+    private func isValidScanlatorMatch(for next: AidokuRunner.Chapter, current: Set<String>) -> Bool {
+        let nextScanlators = Set(next.scanlators ?? [])
+        return current.isEmpty ? nextScanlators.isEmpty : !current.isDisjoint(with: nextScanlators)
+    }
+
     private func findBestChapterMatch(from index: Int, step: Int) -> AidokuRunner.Chapter {
         let firstCandidate = chapterList[index]
         let currentScanlators = Set(chapter.scanlators ?? [])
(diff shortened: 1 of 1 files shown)


kunihir0:
changes are good, just wanted to cut down on additions to the file as it is getting big but build failled with those changes so Ill have to check that rq

iOS/UI/Reader/ReaderViewController.swift
```diff
@@ -589,6 +589,38 @@ extension ReaderViewController {
...
+    private func areDuplicates(_ a: AidokuRunner.Chapter, _ b: AidokuRunner.Chapter) -> Bool {
+        a.chapterNumber == b.chapterNumber
+            && a.volumeNumber == b.volumeNumber
+            && (!(a.chapterNumber == nil && a.volumeNumber == nil) || a.title == b.title)
+    }
+
+    private func findBestChapterMatch(from index: Int, step: Int) -> AidokuRunner.Chapter {
+        let firstCandidate = chapterList[index]
+        let currentScanlators = Set(chapter.scanlators ?? [])
+
```
Skittyblock: just this line with extra spacing; you can turn on trim trailing whitespace for empty lines in xcode preferences

kunihir0:
now it build successfully lol forgot the a but we good now

kunihir0:
bug 1: when reading a source that has no scanlator information for example chapter 1 will have 4 sources, 3 of which are different groups scanlator 1,3,4. we will pick scanlator 2 this source doesnt have any scanlator metadata for us, chapter 2 will have the same source. when we continue from chapter 1 to 2. it will default back to the first in the list which is scanlator 1 instead of it being scanlator 2.

kunihir0:
had to revert the `guard !currentScanlators.isEmpty` check. found a bug where it breaks continuity for sources that don't provide scanlator metadata.

if the current chapter has no scanlators ([]), that guard forces an immediate return of `firstCandidate` this means if i'm reading a version with no metadata, it defaults back to the first available in the list for the next chapter example, switching from "no-group" to "scanlator A", instead of finding the matching "no-group" version.

removed the guard and updated the loop to explicitly match "empty" with "empty" so it sticks to the correct source.

PR #911, opened 2026-02-24
fix: continue reading from last position instead of first chapter

kunihir0:
The "Start/Continue Reading" button previously defaulted to the first uncompleted chapter in the list (based on sort order). This caused issues when multiple source versions of the same chapter existed or when a user skipped chapters, as the button would not resume the specific chapter the user was last reading.

This change now checks for an active reading session first; if none is found, it falls back to the original behavior (first unread chapter in sort order).


Skittyblock (approved):
this looks good, thank you for the pr!

PR #925, opened 2026-03-10
fix(komga): tags display on the `oneShot` type book details page is incomplete

dyphire:
Before:
图片

After:
图片


Shared/Sources/Komga/KomgaModels.swift
```diff
@@ -184,7 +186,7 @@ extension KomgaSearchCondition: Encodable {
         case mediaProfile // books only
         case mediaStatus // books only
         case numberSort // books only
-        case oneShot
+        case oneshot
```
Skittyblock: shouldn't this be `oneShot`? https://github.com/gotson/komga/blob/a584f4cf12ea6344db140732e5873c54c41d4fce/komga/src/main/kotlin/org/gotson/komga/domain/model/SearchCondition.kt#L75

it's never used anyways(?) though so i don't think it needs to be uncommented

PR #931, opened 2026-03-13
feat: improve tags display with expandable view and wrapping layout

### LINKED ISSUE #366: Allow for genres to be expanded
Currently, the genre chips are only scrollable, it's fine for manga with a couple of genres, but there are some that have a lot of genres and scrolling through them is not the best UX. I believe it would be better to allow for the genres to either just be multi line by showing all of them (no more scrolling), or allow for them to be expanded like the descriptions.

dyphire:
Close #366

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -131,6 +131,8 @@
 "BROWSE" = "Browse";
 "BROWSE_NO_SOURCES" = "No Sources";
 "BROWSE_NO_SOURCES_TEXT_NEW" = "Add a source with the plus button.";
+"EXPAND_TAGS" = "Expand";
+"COLLAPSE_TAGS" = "Collapse";
 "ADDING_SOURCES_GUIDE_BUTTON" = "Adding Sources Guide";
 "UPDATES" = "Updates";
 "INSTALLED" = "Installed";
--- a/Shared/Localization/zh-Hans.lproj/Localizable.strings
+++ b/Shared/Localization/zh-Hans.lproj/Localizable.strings
@@ -74,6 +74,8 @@
 
 // Manga
 "UNTITLED" = "无标题";
+"EXPAND_TAGS" = "展开";
+"COLLAPSE_TAGS" = "收起";
 "ONGOING" = "连载中";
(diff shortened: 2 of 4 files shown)


Skittyblock (changes requested):
not really a fan of this as-is
- there's too much extra padding
- having (up to) four lines of tags, combined with four lines of description, takes up too much space and pushes chapters down too far

i personally prefer the single line of tags, but i do acknowledge that it isn't sufficient for entries with a large amount of them. i need to think more about what a ui/x for this would be

dyphire:
> having (up to) four lines of tags, combined with four lines of description, takes up too much space and pushes chapters down too far

I don't think this is a problem, the tags can be collapsed again after it is expanded. The description should also be collapsible, I just didn't implement this.

> i personally prefer the single line of tags, but i do acknowledge that it isn't sufficient for entries with a large amount of them. i need to think more about what a ui/x for this would be

The user experience of single-line horizontal scrolling is really bad. I'm really looking forward to your better designs and ideas.

Skittyblock:
> I don't think this is a problem, the tags can be collapsed again after it is expanded

specifically there I meant in the default unexpanded view. if you have a long description and many tags it pushes the chapters down too far, with no option to collapse further

dyphire:
> > I don't think this is a problem, the tags can be collapsed again after it is expanded
> 
> specifically there I meant in the default unexpanded view. if you have a long description and many tags it pushes the chapters down too far, with no option to collapse further

I see what you mean, and this consideration makes sense.

Let's keep the default display of horizontal scrolling, and only provide an expand button to switch to an expanded layout when scrolling is actually needed.

dyphire:
Any suggestions? I think the current design and implementation are usable enough.

Skittyblock:
i'm still thinking about this and will come back to it before the next update

but the main reasons this current version wouldn't work as-is are:
- the initial scroll view is cut off on the sides
- having the expand button makes a bunch of empty space at the bottom i don't like
- the expanded view can have too much empty space with the line breaks (the space at the end of a line)

also thinking about how to separate artist and author names somehow. maybe we could do a separate metadata page that could open up and display things less compactly, but that might be a bit too out of the way just to view tags so i'm not sure

PR #935, opened 2026-03-18
feat: offset double page spreads and split wide image improvements

### LINKED ISSUE #278: Chapter not properly loading on next
Awesome app, really love the UX / UI choices and it's super lightweight and fast! 

I started using it as my main for reading, however I encountered a pretty annoying bug which initially thought I was daydreaming.

When you finish a chapter and go to the next, only one page loads up (making it seem the beginning of the chapter only has 1 page). However, if you go back to the list and you reopen that exact same chapter then 2 pages load up (as it should be).

I managed to recreate it, since it happens all the time (https://we.tl/t-yv5L2Rzyac). Let me know if you also need the logs.

Thanks for your hard work!

### LINKED ISSUE #445: Ability to offset spreads

### LINKED ISSUE #745: Split Page View not functioning properly when using upscaler.
### Summary

When using any upscaler, the page will not split the page properly. In my case, it only split a small portion of the image into 2 pages. This happens with both the Waifu2x and Real-Esgran models. Waifu2x is not loading correctly for me but I can tell it is not the full page. 

### Steps to reproduce

1. Enable “Split WIde Images” in Reader settings.
2. Enable an upscaler in the Upscaling section.
3. Open a wide page that should be split. It will not be correctly split.

It seems that the upscaler and split page function are colliding and the image is not being split correctly.
Original Image: 3040 x 2400
Split Image: 1520 x 2400
Upscaled Image: 6080 x 4800
Upscaled Split Image: 1014 x 1600

The image is upscaled and then split.

### Screenshots

I paused the recording during the upscaling which is why it appears to upscale so quickly. 

https://github.com/user-attachments/assets/5ebd2284-2a29-4c3d-8fc3-aea1b69c9ffd

### Device

iPhone 16 Pro Max

### Operating System

iOS

### OS Version

26.0.1

### App Version

0.7.2

### Acknowledgements

### LINKED ISSUE #918: Flashing when pre buffering page  in split view
### Summary

When a single comic image contains 2 pages. i use split view (spilt wide page) , then it will flashing. not comfortable for eyes. please fix it , thanks.

### Steps to reproduce

Open a two pages in a single image comics ,and make spilt wide enabled.

### Device and version

iPhone 17 Pro Max

### LINKED ISSUE #939: Wide images (spreads) incorrectly forced into double-page layout, resulting in black screen
### Summary

On iPad mini 6, when "Automatic Single/Double Page" is enabled, the app fails to recognize a wide image (landscape spread) as a single page. Instead, it treats it as a double-page container. In landscape orientation, this causes a rendering failure where the viewer displays a total black screen.

### Steps to reproduce

Device: iPad mini 6 (Landscape).
Settings: Viewer Mode -> Automatic Single/Double Page.
Action: Open a manga with wide/landscape images (spreads).
Result: The wide image is not rendered; the screen is black.
Expected behavior 
A wide image (Landscape orientation) should be treated as a Single Page (Spread) that fills the screen. It should NOT be processed by the double-page pairing logic

### Device and version

Device: iPad mini 6   iOS Version: ( iOS 26.3.1) Aidoku Version: (0.8.2)

chibijoam:
may also have something I didn’t notice, needs testing

closes #445 
closes #745
closes #278


ghost:
all issues i know are fixed

ghost:
behavior needs to be corrected

ghost:
all issues i know are fixed :)

Skittyblock (changes requested):
thank you very much for this, i've tested and it seems to work well, and even seems to fix #918 which is great.

the only other comment i have is on the functionality of the setting/toggling. as it stands, i don't think it's very intuitive that you need to first turn on the setting and then toggle with `o` or the double tap. what would be better, imo, is if the toggling always worked, and if it toggled the setting directly. turning the setting on and off in the settings page should do the same thing as the key/gesture.

also, is the "always isolate first page" setting still useful with this? i think it can probably be removed since it was just a placeholder for this more complex setting

ghost:
> thank you very much for this, i've tested and it seems to work well, and even seems to fix #918 which is great.
> 
> the only other comment i have is on the functionality of the setting/toggling. as it stands, i don't think it's very intuitive that you need to first turn on the setting and then toggle with `o` or the double tap. what would be better, imo, is if the toggling always worked, and if it toggled the setting directly. turning the setting on and off in the settings page should do the same thing as the key/gesture.
> 
> also, is the "always isolate first page" setting still useful with this? i think it can probably be removed since it was just a placeholder for this more complex setting

you can retest it :D

ghost:
closes #745

found some issues with this fix — may cause page index drift when using split wide images, currently fixing

ghost:
> closes #745
> 
> found some issues with this fix — may cause page index drift when using split wide images, currently fixing

all issues i know are fixed

likely also fixes #278

iOS/UI/Reader/Page/ReaderPageView.swift
```diff
@@ -499,6 +512,36 @@ class ReaderPageView: UIView {
             imageAnalaysisInteraction?.isSupplementaryInterfaceHidden = hidden
         }
     }
+
+    // use Task to run LiveText analysis to prevent wide image jumps
+    private func startLiveTextAnalysis() {
+        if #available(iOS 16.0, *) {
+            guard imageAnalaysisInteraction != nil, let image = imageView.image else { return }
```
Skittyblock: probably better to include this as one of the guard conditions instead of wrapping the whole function body in an if block
```swift
        guard
            #available(iOS 16.0, *),
            imageAnalaysisInteraction != nil,
            let image = imageView.image
        else {
            return
        }
```

Skittyblock (changes requested):
this works great! i'm excited at how many issues this fixes...

there are two extra issues i've found, but they're not super important, and i could fix them later:

- the reload button no longer shows in double pages if the page fails to load. i think this was finicky before as well though
- there's a crash with text pages in the double page controller due to how the view is moved into the stack view with an incorrect parent, but this issue was already present in the main branch i think
```
*** Terminating app due to uncaught exception 'UIViewControllerHierarchyInconsistency', reason: 'child view controller:<_TtGC7SwiftUI19UIHostingControllerV6Aidoku12MarkdownView_: 0x10a292200> should have parent view controller:<Aidoku.ReaderDoublePageViewController: 0x10a292e00> but actual parent is:<Aidoku.ReaderPageViewController: 0x10a255200>'
```

both of these can be seen if you uncomment the Demo Source code in `AddSourceView.swift` to add and read from.

Skittyblock (approved):

PR #947, opened 2026-03-28
Resume most recently opened chapter from Start Reading

phonixdrive:
Start Reading currently ignores the last chapter you actually opened if that chapter is already marked as read, which can make it jump to the wrong place after skipping around.

This changes it to prefer the most recently opened chapter from reading history first, and only fall back to the first unread chapter if there’s no history yet.

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -431,6 +431,7 @@
 "LANDSCAPE" = "Landscape";
 "READER" = "Reader";
 "OPEN_READER_VIEW" = "Open Reader View";
+"RESUME_LAST_OPENED_CHAPTER" = "Resume Last Opened Chapter";
 "UNREAD_CHAPTER_BADGES" = "Unread Chapter Badges";
 "DOWNLOADED_CHAPTER_BADGES" = "Downloaded Chapter Badges";
 "PIN_TITLES" = "Pin Titles";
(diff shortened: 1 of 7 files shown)


Skittyblock (changes requested):
it looks like this removes the functionality where it will default to opening the last read chapter if it's not completed. it should still do this if the new setting is off

phonixdrive:
It should be fixed now. It should use the original logic if the setting is set to false.

Skittyblock:
> It should use the original logic if the setting is set to false.

you removed the `Find the most recently read chapter that isn't completed` part (replacing it with the new setting), so if the new setting is off it won't do that

phonixdrive:
Reverted and just separated into its own function.

Skittyblock (changes requested):
i think integrating it in the `getNextChapter` function instead of separately would be more efficient, since then you only need to iterate through the chapter list once in the case of there being no opened chapters. you only need to conditionally check if `history.page != -1` when the setting is off

also, the function is called in two places: with the continue reading button, and when opening a manga in the library with the "Open Reader View" setting on, and this change looks like it doesn't have any effect on the library opening

Skittyblock (changes requested):
final two comments, should be good after this

Skittyblock (approved):

phonixdrive:
Changes look good, should work.

PR #951, opened 2026-03-29
fix: address text reader issues

rbertus2000:
- Fix word truncation by matching lineFragmentPadding between paginator and UITextView
- Fix content shifting in scroll reader when bars toggle via safe area compensation
- Enable tap zones by disabling UITextView user interaction on read-only pages
- Add two-step keyboard navigation at chapter boundaries (transition page first)
- Preserve read position when switching between paged and scroll reader styles
- Fix sentence break edge case that could split words across pages

### DIFF
--- a/iOS/Extensions/HostingController.swift
+++ b/iOS/Extensions/HostingController.swift
@@ -8,9 +8,15 @@
 import SwiftUI
 
 final class HostingController<Content: View>: UIHostingController<Content> {
+    /// When true, skips `invalidateIntrinsicContentSize` during layout.
+    /// Used by the scroll text reader to prevent size recalculation during bar transitions.
+    var suppressInvalidation = false
+
     override func viewDidLayoutSubviews() {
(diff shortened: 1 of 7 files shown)


Skittyblock (changes requested):
it seems like the fix for the scroll reader shifting isn't very consistent; it doesn't work at the top of a chapter, or if i scroll on to a second text chapter

the other fixes appear to work fine

rbertus2000:
The scroll reader now shouldn't jump in iOS >=16.4. In older versions it is at least reduced.

Skittyblock (changes requested):
this does indeed work to fix the shifting, but now it looks like it starts chapters with the text covered by the navbar, which it didn't do previously. is it possible to fix that?

rbertus2000:
It should be fixed now.

Skittyblock (approved):
seems to work good now, thanks

PR #952, opened 2026-03-31
feat: sort available sources when migrating manga

Adamekka:
Sorts available sources when migrating manga to a new source

Screenshot 2026-03-31 at 9 21 56 PM

### DIFF
--- a/iOS/UI/Migration/MigrateSourceSelectionView.swift
+++ b/iOS/UI/Migration/MigrateSourceSelectionView.swift
@@ -12,7 +12,9 @@ struct MigrateSourceSelectionView: View {
     var excludedSources: [String] = []
 
     @Binding var selectedSources: [SourceInfo2]
-    @State var availableSources = SourceManager.shared.sources.map { $0.toInfo() }
+    @State var availableSources = SourceManager.shared.sources
+        .map { $0.toInfo() }
+        .sorted { $0.name.localizedCaseInsensitiveCompare($1.name) == .orderedAscending }
 
     @State private var editMode: EditMode = .active
 


iOS/UI/Migration/MigrateSourceSelectionView.swift
```diff
@@ -12,7 +12,9 @@ struct MigrateSourceSelectionView: View {
     var excludedSources: [String] = []
 
     @Binding var selectedSources: [SourceInfo2]
-    @State var availableSources = SourceManager.shared.sources.map { $0.toInfo() }
+    @State var availableSources = SourceManager.shared.sources
+                                    .map { $0.toInfo() }
+                                    .sorted { $0.name.localizedCaseInsensitiveCompare($1.name) == .orderedAscending }
```
Skittyblock: we prefer not to try and do any alignment, so it would be better like this
```suggestion
    @State var availableSources = SourceManager.shared.sources
        .map { $0.toInfo() }
        .sorted { $0.name.localizedCaseInsensitiveCompare($1.name) == .orderedAscending }
```

Skittyblock (changes requested):
thank you for the pr

Adamekka:
I squashed the changes

Skittyblock (approved):

Adamekka:
This PR has one flaw. I didn't know that the sources were already sorted, but based on language.

I tried making the Migration View same as the list of sources, but I stumbled on iOS BS you already were hacking, so it doesn't look good.

Tell me if I should open a new PR with this, even if it's tinted and doesn't look the best.

Setting the button style to plain fixes the tint, but breaks the hitbox of the button, so only the text is selectable.

image

Skittyblock:
i pushed a68c533f590ccab35d194f768193c9c7cc475c18 to add icons and revert the sorting change. i should've realized that we were already sorting sources the same as their order in the browse tab

PR #956, opened 2026-04-09
fix: reuse hosting controllers in CollectionView cells

chibijoam:


iOS/New/Views/Common/CollectionView.swift
```diff
@@ -53,24 +53,12 @@ struct CollectionView: UIViewControllerRepresentable {
...
-                hostingController.view.bottomAnchor.constraint(equalTo: cell.contentView.bottomAnchor),
-                hostingController.view.leadingAnchor.constraint(equalTo: cell.contentView.leadingAnchor),
-                hostingController.view.trailingAnchor.constraint(equalTo: cell.contentView.trailingAnchor)
-            ])
-
+            guard
+                let cell = collectionView.dequeueReusableCell(withReuseIdentifier: "Cell", for: indexPath)
+                    as? UIHostingCollectionViewCell<AnyView>
+            else { return UICollectionViewCell() }
+            // there's a bug on ios <=16 where the views will shift down after you scroll down
```
Skittyblock: this comment was just about why `ignoreSafeArea` was used i think, but it seems to not be an issue with `UIHostingCollectionViewCell` (?)

Skittyblock (approved):

PR #985, opened 2026-05-22
feat: notifications for new chapters in background

### LINKED ISSUE #94: Notifications for newly released chapters
It would be nice if Aidoku could update the chapter list of everything in your library every so often and send a notification when it detects a new chapter has been added. Don't know if this is possible/feasible on iOS but if it is it'd be really nice.

Zareix:
Closes #94 .
This PR adds the notification for when a new chapter is available.
This is based on the background refresh system of library that already exists .

Tested on my iPhone and working.


Shared/Managers/NotificationManager.swift
```diff
@@ -0,0 +1,76 @@
...
+
+            try? await center.add(request)
+        }
+    }
+
+    private static func body(for summary: NewChaptersSummary) -> String {
+        if summary.chapterCount == 1 {
+            return NSLocalizedString("1_NEW_CHAPTER_AVAILABLE", comment: "")
+        }
+        return String(format: NSLocalizedString("X_NEW_CHAPTERS_AVAILABLE", comment: ""), summary.chapterCount)
```
Skittyblock:
```suggestion
            return NSLocalizedString("1_NEW_CHAPTER_AVAILABLE")
        }
        return String(format: NSLocalizedString("X_NEW_CHAPTERS_AVAILABLE"), summary.chapterCount)
```

Shared/Managers/NotificationManager.swift
```diff
@@ -0,0 +1,76 @@
...
+
+import Foundation
+import UserNotifications
+
+actor NotificationManager {
+    static let shared = NotificationManager()
+
+    struct NewChaptersSummary {
+        let sourceId: String
+        let mangaId: String
```
Skittyblock: this could be `MangaIdentifier`

Shared/Managers/NotificationManager.swift
```diff
@@ -0,0 +1,76 @@
...
+    func notifyNewChapters(_ summaries: [NewChaptersSummary]) async {
+        guard !summaries.isEmpty, isEnabled() else { return }
+
+        let center = UNUserNotificationCenter.current()
+        let settings = await center.notificationSettings()
+        guard settings.authorizationStatus == .authorized
+            || settings.authorizationStatus == .provisional
+        else { return }
+
+        for summary in summaries {
```
Skittyblock, line 61: does this send a notification for every single series that gets updated? i'm not sure about that...

Skittyblock (changes requested):
i haven't tested this myself, but i assume it's sending notifications while the app is actually running, right? or if a user has triggered a background refresh and left the app. i'm not sure if that's really when the app should be sending notifications. i feel like it should generally only happen if a refresh happens that the user didn't trigger manually (e.g. refreshing an email app doesn't send you notifications when you're inside the app i don't think)

Zareix:
I'm not sure if manual refresh trigger this here, I don't think so but I will double check. 
But the idea here is that I added the trigger in the background refresh function of Aidoku.
So let's say I set it to every 12h. I close the app and in 12h the refresh will start in the background while app is still closed, then if a series has a new chapter then it will send a notification.

Shared/Managers/NotificationManager.swift
Zareix: Indeed it does, I don't have a big library so I didn't think about it.  
However I hardly see, even on big libraries, the app sending more than 3 to 4 update notifications at the same time.
But we could work on sending a batch one if you think it's worth it

Zareix: Something like : 
If there is more than 3 then send a batch one like "X new series with new chapters"
Else send each one alone

Zareix:
After double-checking, it's indeed also called when manually refreshing, so I added a flag to distinguish both and only send notifications on background refresh.

Shared/Managers/NotificationManager.swift
Zareix: I've implemented the above

Skittyblock (changes requested):
i guess i forgot to mention this previously, but i don't think localizations should be included for languages you don't know. it's better to just let translators on weblate handle it than to machine translate it

Skittyblock (approved):

PR #976, opened 2026-05-17
feat: tracker chapter number offset

bcmeireles:
Fixes #910 

Adds a per-title chapter offset for tracker sync to handle source/tracker chapter numbering mismatches (e.g. prologues/bonus chapters)

This was made by adding `Advanced` -> `Set Chapter Offset` to the tracker menu.

Kept manual tracker edits unchanged (offset is not applied there).

Being portuguese I also took the freedom to add localization for pt.

I added `chapterOffset` under a new `0.8.3.xcdatamodel` and updated needed files accordingly. If there is a better approach to this added field please let me know.

The tests ran successfully and I've also tested this manually through an iOS simulator.

Some screenshots for reference:
image
image

PR #988, opened 2026-05-24
fix: recover library covers when source URL goes stale

### LINKED ISSUE #326: Refresh library covers option
Multiple users are being faced with issues with sources that change cover image urls, so the library covers fail to load. A quick option in settings to fix this would be ideal.

neldra:
Fixes #326.

On HTTP 403/404/410 from a library cover load, refetch manga details via `MangaManager.resetCover` and retry with the fresh URL. Throttled to one attempt per manga per session. Applied to both `MangaGridCell` and `MangaListCell`.

Also includes a small unrelated fix updating `TrackerSyncTests` for the `trackItem` rename in #976, so the test target compiles locally.

### DIFF
--- a/iOS/New/Views/Common/MangaListCell.swift
+++ b/iOS/New/Views/Common/MangaListCell.swift
@@ -12,6 +12,7 @@ import UIKit
 
 class MangaListCell: UICollectionViewCell {
     private var sourceId: String?
+    private var mangaId: String?
     private var url: String?
     private var imageTask: ImageTask?
 
@@ -236,6 +237,7 @@ extension MangaListCell {
 extension MangaListCell {
     func configure(with manga: AidokuRunner.Manga, isBookmarked: Bool = false) {
         sourceId = manga.sourceKey
+        mangaId = manga.key
         titleLabel.text = manga.title
         subtitleLabel.text = manga.authors?.joined(separator: ", ")
         subtitleLabel.isHidden = subtitleLabel.text?.isEmpty ?? true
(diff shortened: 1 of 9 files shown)


iOS/UI/Common/Manga/CoverRecovery.swift
```diff
@@ -0,0 +1,44 @@
...
+
+import AidokuRunner
+import Foundation
+import Nuke
+
+enum CoverRecovery {
+    private static let staleStatusCodes: Set<Int> = [403, 404, 410]
+
+    private static var attempted = Set<String>()
+    private static let lock = DispatchQueue(label: "app.aidoku.coverRecovery")
```
Skittyblock: i don't think gcd should be used anymore if we can help it; i think this object (either `CoverRecovery` or a new class just to manage the "attempted" set) could be an `actor` instead

iOS/New/Views/Common/MangaListCell.swift
```diff
@@ -12,6 +12,7 @@ import UIKit
 
 class MangaListCell: UICollectionViewCell {
     private var sourceId: String?
+    private var mangaId: String?
```
Skittyblock: instead of storing both the source and manga ids in separate variables we could use the `MangaIdentifier` struct instead. this is a thing i need to use in the rest of the app since lots of old code separates the two variables. if it's too annoying to change though you could leave it and i'll just go in later and refactor

Skittyblock (changes requested):
thanks, i haven't tested but i think this is probably a good idea to have

neldra:
Addressed both: swapped the GCD lock for a private actor managing the attempt set, and switched the cells + recover over to MangaIdentifier. Thanks for the review!

Skittyblock (approved):
lgtm, will merge before the next app update

PR #992, opened 2026-06-05
Retry manga panel downloads on rate limits or network errors

pxseu:
For a long time I've had a dozen of issues where I wanted to download a manga chapter but the actual downloaded content came out corrupted. After digging deeper I found out that raw html responses were saved (CF-429, etc.), as-is, breaking the reader.

Basically, the downloader will attempt to wait out all rate limits (`Retry-After` header), and then as a fallback I added a really tiny exponential backoff, that can be configured by two new statics.

This also adds some logging so if any future issues arise this should be simple to fix =]

I don't usually write swift, so I would like to apologise if anything looks messy.

### DIFF
--- a/Shared/Data/Downloads/DownloadTask.swift
+++ b/Shared/Data/Downloads/DownloadTask.swift
@@ -36,6 +36,10 @@ actor DownloadTask: Identifiable {
 
     private static let maxConcurrentPageTasks = 5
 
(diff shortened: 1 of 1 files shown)


Skittyblock (approved):
this looks fine to me, but i'll do a little testing before merging before the next update. thank you for the contribution!

PR #1001, opened 2026-06-15
feat: optimize backup restore time

Amqx:
use a dictionary with lookup keys instead of repeated linear scans

this speeds up restore times massively

tested on:
1184 library entries
62906 history entries
128501 chapters

before took 17 minutes to fully restore, now takes 10 seconds

### DIFF
--- a/Shared/Data/Backup/BackupManager.swift
+++ b/Shared/Data/Backup/BackupManager.swift
@@ -12,48 +12,6 @@ import Foundation
 import UIKit
 #endif
 
-private extension BackupLibraryManga {
-    var identifier: MangaIdentifier {
-        .init(sourceKey: sourceId, mangaKey: mangaId)
-    }
-}
-
-private extension BackupChapter {
-    var identifier: ChapterIdentifier {
-        .init(sourceKey: sourceId, mangaKey: mangaId, chapterKey: id)
-    }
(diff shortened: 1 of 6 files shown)


Shared/Data/Backup/BackupManager.swift
```diff
@@ -12,6 +12,72 @@ import Foundation
 import UIKit
 #endif
 
+struct BackupRestoreMangaKey: Hashable, Sendable {
```
Skittyblock: could you use `MangaIdentifier` and `ChapterIdentifier` instead? they're the same thing
`MangaObject` and stuff also have the `identifier` property that could be used instead of creating new initializers

Skittyblock (changes requested):
dang thank you, i've been meaning to revisit this. feel dumb for not realizing that was the bottleneck, for some reason i just assumed it was the coredata calls

Amqx:
I'll look into it when I get home tn

Shared/Data/Backup/BackupManager.swift
Amqx: modified to use the identifiers

Amqx:
i think i got what you wanted

Skittyblock (approved):

PR #1017, opened 2026-07-05
feat: add initial epub support

### LINKED ISSUE #609: epub support
i'm pivoting aidoku to a more general content reading rather than strictly manga, so epub support would be beneficial to have for both komga servers and local file reading.

rbertus2000:
## Summary

Adds EPUB support to the local source. EPUB files can now be imported and read with the existing text readers (paged and scroll), including inline images.

Closes #609 

## How it works

- **Import:** One EPUB file = one series. `EpubParser` reads the OPF spine and groups spine files into logical chapters based on the TOC (`toc.ncx` / nav document), so multi-file chapters appear as a single chapter entry.
- **Chapter IDs** use the format `<epub file name>/<primary href>`, decoded in `LocalFileManager.readEpubPages`.
- **Text chapters** are converted into a single markdown `.text` page with inline `image` references; images are extracted to `Caches/EpubImages/`.
- **Image-only chapters** (cover pages, illustration galleries) are kept as `.zipFile` pages and open in the regular image reader.
- Small decorative images (<100 KB, e.g. chapter ornaments) stay inline in text chapters instead of forcing the chapter into the image reader.

## Reader changes

- `TextPaginator` renders inline images via `NSTextAttachment`, capped to the content size so oversized images can't swallow the remaining page text.
- `MarkdownView` (scroll reader) gets a `LocalFileImageProvider`, since MarkdownUI's default provider ignores `file://` URLs.

## Also included

- fix(textreader): align double-page spreads to even page indices
- fix(textreader): `alignedPageIndex` never reported a finished chapter

## Testing

- Imported EPUBs (incl. J-Novel Club light novels with per-heading ornament images and illustration inserts) and verified chapter grouping, inline images, and progress tracking in both paged and scroll text readers.

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -217,7 +217,7 @@
 "IMPORT" = "Import";
 "IMPORT_FILE" = "Import File";
 "LOCAL_FILE_IMPORT" = "Local File Import";
-"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ, ePub or ZIP file containing images to read.";
+"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ, ePub, or ZIP file containing images to read.";
 "CHAPTER_TITLE" = "Chapter Title";
 "SERIES_TITLE" = "Series Title";
 "DESCRIPTION" = "Description";
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -217,7 +217,7 @@
 "IMPORT" = "Import";
 "IMPORT_FILE" = "Import File";
 "LOCAL_FILE_IMPORT" = "Local File Import";
-"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ or ZIP file containing images to read.";
+"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ, ePub or ZIP file containing images to read.";
 "CHAPTER_TITLE" = "Chapter Title";
 "SERIES_TITLE" = "Series Title";
 "DESCRIPTION" = "Description";
@@ -235,6 +235,7 @@
 "1_PAGE" = "1 page";
 "CBZ_NAME" = "Comic Book Zip Archive";
 "ZIP_NAME" = "Zip Archive";
+"EPUB_NAME" = "EPUB";
 
 // Komga
 "KOMGA" = "Komga";
--- a/Shared/Sources/Local/LocalFileManager.swift
+++ b/Shared/Sources/Local/LocalFileManager.swift
@@ -21,7 +21,7 @@ actor LocalFileManager {
     private var lastScanTime = Date.distantPast
     private var scanTask: Task<Void, Never>?
 
-    static let allowedFileExtensions = Set(["cbz", "zip"])
+    static let allowedFileExtensions = Set(["cbz", "zip", "epub"])
     static let allowedImageExtensions = Set(["jpg", "jpeg", "png", "webp", "gif", "heic", "avif"])
     static let allowedTextExtensions = Set(["txt", "md"])
     static let allowedPageExtensions = allowedImageExtensions.union(allowedTextExtensions)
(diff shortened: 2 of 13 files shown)


Shared/Localization/en.lproj/Localizable.strings
```diff
@@ -217,7 +217,7 @@
 "IMPORT" = "Import";
 "IMPORT_FILE" = "Import File";
 "LOCAL_FILE_IMPORT" = "Local File Import";
-"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ or ZIP file containing images to read.";
+"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ, ePub or ZIP file containing images to read.";
```
Skittyblock: should use an oxford comma
```suggestion
"LOCAL_FILE_IMPORT_TEXT" = "Import a CBZ, ePub, or ZIP file containing images to read.";
```

Shared/Sources/Local/LocalFileManager.swift
```diff
@@ -131,9 +150,84 @@ extension LocalFileManager {
...
+    // be referenced with a file url from markdown text
+    nonisolated static func cacheEpubImage(archiveURL: URL, path: String) -> URL? {
+        guard let cachesDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first
+        else { return nil }
+
+        // stable per-book folder name (fnv-1a hash of the archive path)
+        var hash: UInt64 = 0xcbf29ce484222325
+        for byte in archiveURL.path.utf8 {
+            hash = (hash ^ UInt64(byte)) &* 0x100000001b3
+        }
```
Skittyblock: should probably be a function (can just be inline)
```swift
// stable per-book folder name (fnv-1a hash)
func hash(_ s: String) -> String {
    var hash: UInt64 = 0xcbf29ce484222325
    for byte in s.utf8 {
        hash = (hash ^ UInt64(byte)) &* 0x100000001b3
    }
    return String(hash, radix: 16)
}
hash(archiveURL.path)
```

Shared/Sources/Local/LocalFileManager.swift
```diff
@@ -131,9 +150,84 @@ extension LocalFileManager {
...
+        // image-only chapter (cover, illustration pages)
+        return segments.compactMap { segment in
+            guard case let .image(path) = segment else { return nil }
+            return AidokuRunner.Page(content: .zipFile(url: archiveURL, filePath: path))
+        }
+    }
+
+    // extract an image from an epub archive into the cache directory so it can
+    // be referenced with a file url from markdown text
+    nonisolated static func cacheEpubImage(archiveURL: URL, path: String) -> URL? {
```
Skittyblock: it would be nice if we could hijack nuke's disk image cache, but i'm not sure if that's possible. maintaining a separate cache seems annoying; does this ever get cleared?

Skittyblock (changes requested):
haven't tested yet or looked super closely but this seems reasonable for initial epub support, thank you!
with this we should also be able to support epubs from kavita/komga, but i can make the necessary changes for that

Shared/Sources/Local/LocalFileManager.swift
rbertus2000: The cache lives in Caches/, so the system can purge it under storage pressure, but it wasn't explicitly cleared before — fixed now: it's cleaned up in removeManga/removeChapter/removeAllLocalFiles. Reusing Nuke's disk cache doesn't really work here since the markdown text needs stable file:// URLs that both NSTextAttachment and MarkdownUI can resolve, and Nuke's DataCache doesn't expose stable file paths as public API.

Skittyblock (changes requested):
after testing the only real issue i found is that removing one chapter will remove the epub, causing the rest of the chapters to stop working. it should either remove all chapters associated with that epub, or only remove the epub file once all chapters have been removed from it.

i also found that importing the same epub twice and having it create a new series with the same name will have some odd behavior: it will import the existing series, but create two entries in the list where you select which series to import to. but i think this was an existing issue that was also present with cbzs so it doesn't need to be addressed in this pr

rbertus2000:
> after testing the only real issue i found is that removing one chapter will remove the epub, causing the rest of the chapters to stop working. it should either remove all chapters associated with that epub, or only remove the epub file once all chapters have been removed from it.
> 
> i also found that importing the same epub twice and having it create a new series with the same name will have some odd behavior: it will import the existing series, but create two entries in the list where you select which series to import to. but i think this was an existing issue that was also present with cbzs so it doesn't need to be addressed in this pr

I fixed it. It will now only remove the epub file once all chapters have been removed.

Skittyblock (approved):

PR #1023, opened 2026-07-10
feat: add ocr dictionary lookup

GameFuzzy:
Adds OCR-backed dictionary lookup to image reader pages.

- Adds hoshidicts-backed dictionary lookup and multi-dictionary import support.
- Adds tap and long-press lookup flows for OCR text on paged and webtoon readers.
- Adds dictionary popup rendering based on Hoshi Reader behavior.
- Adds reader dictionary settings.


Skittyblock:
code mostly cleaned up now, just needs more testing. there's other stuff i want to do (e.g. vocab saving), but i'll probably leave that to a later change after this is merged

Skittyblock:
ok now all features i wanted are done, so i will go back to testing a bit before merging. anki and audio (and image context for saved words) can come later

might do some code cleanup of the stuff from hoshi reader than aidoku doesn't need

Skittyblock:
should be good enough for merge now. main issue is just the unreliability of the ocr for some words, however i've compared it with live text and it looks the same, so afaik this is just the limit of the vision framework. a better solution would probably be using a dedicated ocr library (for japanese text in particular), however i'd be apprehensive about adding any large dependencies that could significantly increase the binary size when this is a feature i assume the majority of users have no interest in. i'll look into it at some point in the future if there is interest, but this is good enough for the initial feature support

PR #1027, opened 2026-07-11
feat: customizable text reader colors

rbertus2000:
## Summary

This PR adds preset color themes to the text reader (Books-style, as suggested in the review), selectable separately for light and dark mode, and brings the liquid glass top edge effect (already present in the paged and webtoon image readers) to both text reader modes.

## Features

**Color themes**
- Six presets: **System**, **White**, **Sepia**, **Paper** (light gray), **Gray** (dark gray), and **Black** — no free color selection.
- Text color is derived from the theme: pure black/white on white/black backgrounds, slightly transparent black/white on the tinted ones.
- Themes other than System override the reader's interface style (e.g. Sepia forces light, Gray forces dark), so the bars, transition pages, and navigation title match the page background.
- The theme can be chosen separately for light and dark appearance (both default to System), so e.g. Sepia during the day and Gray at night switch automatically with the device theme — generalizing the suggested "system" option.
- Configurable in the in-reader settings sheet and the global reader settings.

**Liquid glass top edge effect**
- Both text reader modes apply the soft scroll edge effect at the top edge on iOS 26+, matching the image readers, tied to bar visibility so it doesn't dim the page while reading immersively.

## Implementation notes

- `ReaderTextTheme` enum provides the palette, per-theme text color, and interface style; stored as raw values under `Reader.textThemeLight`/`Reader.textThemeDark`. The active slot follows the window's interface style (not the reader's, which may be overridden by the theme itself).
- Device light/dark flips while reading are picked up via `registerForTraitChanges` on the window (iOS 17+).
- The navigation bar title color is written into the bar's appearance objects, since those take precedence over both the trait override and `titleTextAttributes`.
- Settings use a plain select via the existing `SettingView`; localized strings reuse the existing `READER_BG_COLOR_*` keys where possible, new strings added for English and German.

## Testing

- [x] Themes apply in both text reader modes and update live when changed
- [x] Light/dark slots switch automatically with the device appearance
- [x] Non-system themes override the reader appearance (bars, title, transition pages)
- [x] Top edge effect appears with bars visible and hides with the overlay
- [x] Image readers unaffected

## Screenshots

| Sepia | Gray | Settings |
| --- | --- | --- |
| Mittel (Screenshot iPhone 17 24 07 2026 at 21 49 21) | Mittel (Screenshot iPhone 17 24 07 2026 at 21 49 29)| Mittel (Screenshot iPhone 17 24 07 2026 at 21 49 43) |


Skittyblock (changes requested):
sorry for the late review

i think making colors *completely* customizable is going farther than we should, making it more customizable at the cost of making it more complex than most people would care for. for example, i don't think there's any reason someone would ever want a neon green background with pink text

personally, i think in the future we should have a theme system more like the books or kindle apps in. it's not necessary to go this far initially, but instead of allowing any color to be set, we should have:
- only a few options for page background color: white, black, sepia/tan, maybe a dark gray and light gray (books app "quiet" and "paper" themes)
  - probably have a "system" option as well that's just black/white depending on device theme. all other page background colors should override the reader's user interface style
- font color should match the theme (white or black on black or white backgrounds, or a slightly transparent black or white on different color backgrounds)

then in a later change we could make it so font, spacing, padding, and colors can all be configured as a part of a theme (don't worry about it for now).

rbertus2000:
> sorry for the late review
> 
> 
> 
> i think making colors *completely* customizable is going farther than we should, making it more customizable at the cost of making it more complex than most people would care for. for example, i don't think there's any reason someone would ever want a neon green background with pink text
> 
> 
> 
> personally, i think in the future we should have a theme system more like the books or kindle apps in. it's not necessary to go this far initially, but instead of allowing any color to be set, we should have:
> 
> - only a few options for page background color: white, black, sepia/tan, maybe a dark gray and light gray (books app "quiet" and "paper" themes)
> 
>   - probably have a "system" option as well that's just black/white depending on device theme. all other page background colors should override the reader's user interface style
> 
> - font color should match the theme (white or black on black or white backgrounds, or a slightly transparent black or white on different color backgrounds)
> 
> 
> 
> then in a later change we could make it so font, spacing, padding, and colors can all be configured as a part of a theme (don't worry about it for now).

So just to get this right:
You want me to remove the font color option and change the background color options to a few like system, black, sepia, grey, paper with corresponding font colors?

Skittyblock:
yes, unless you want to suggest a different solution

rbertus2000:
Reworked as suggested: replaced the color pickers with preset themes (system/white/sepia/paper/gray/black). I also updated the original PR text and pictures.

iOS/New/Views/Reader/ReaderSettingsView.swift
```diff
@@ -244,6 +244,28 @@ struct ReaderSettingsView: View {
...
+                                value: .select(.init(
+                                    values: ReaderTextTheme.allCases.map(\.rawValue),
+                                    titles: Self.textThemeTitles
+                                ))
+                            )
+                        )
+                        SettingView(
+                            setting: .init(
+                                key: ReaderTextTheme.darkUserDefaultsKey,
+                                title: NSLocalizedString("TEXT_THEME_DARK"),
```
Skittyblock, line 261: i don't think the themes should be separated based on system theme. if necessary, a better solution would be to have a separate setting for the system/light/dark appearance, and then making all the themes work with either rather than being locked to one of them

iOS/New/Views/Reader/ReaderSettingsView.swift
```diff
@@ -401,3 +423,15 @@ struct ReaderSettingsView: View {
         }
     }
 }
+
+extension ReaderSettingsView {
+    /// Localized titles for the text reader themes, in `ReaderTextTheme.allCases` order.
```
Skittyblock: these should be on `ReaderTextTheme` itself, not here

iOS/UI/Reader/Readers/Text/Scroll/ReaderTextViewController.swift
```diff
@@ -200,6 +204,20 @@ class ReaderTextViewController: BaseViewController {
...
+            self, selector: #selector(showTopEdgeEffect),
+            name: .readerShowingBars, object: nil
+        )
+        NotificationCenter.default.addObserver(
+            self, selector: #selector(hideTopEdgeEffect),
+            name: .readerHidingBars, object: nil
+        )
+        if navigationController?.navigationBar.isHidden == true {
+            hideTopEdgeEffect()
+        }
```
Skittyblock: this shouldn't be necessary after dc7164d9e9764606a227e4d3dce1c2bb1fc1da16, i realized i accidentally committed to the dictionary pr instead of main

iOS/UI/Reader/Readers/Text/Scroll/ReaderTextView.swift
```diff
@@ -16,6 +16,8 @@ struct ReaderTextView: View {
    let fontSize: Double
    let lineSpacing: Double
    let horizontalPadding: Double
+   // stored (not read in body) so theme changes invalidate the view
+   let textColor = Color(uiColor: ReaderTextTheme.text)
```
Skittyblock: is that how it works..? this variable is never changed so it wouldn't invalidate the view
but it *is* read in the body, so i'm confused by the comment

Skittyblock (changes requested):

iOS/New/Views/Reader/ReaderSettingsView.swift
rbertus2000: I don't think I understand. There are the same available options for light and dark mode. It's just that you can choose the color per mode, especially for people like me who use the auto setting for light and dark mode.

rbertus2000:
I have addressed your feedback. 
Regarding the light/dark theme settings I think I see what you mean now — one theme selection plus a separate appearance setting, with each theme having a light and dark variant (like Apple Books does it).

My concern with that: it doubles the palette (a dark sepia, a light "quiet" gray, …) and takes away the combination I actually built this for — e.g. sepia in light mode but pure black in dark mode. With the two selects, both default to "system", so users who don't care never see a difference, and users on auto appearance get exactly the theme they want per mode.

Skittyblock (changes requested):
what i meant was that there shouldn't be a separate option for light and dark themes. it should only be a single theme option

and then additionally (optionally) an override for the theme appearance 
- if this exists then the "white" and "black" themes would not need to exist, and system could just be "default". overriding the theme appearance should also override the reader controller interface style
- if this doesn't exist, the themes should have an overridden interface style. e.g. if you select the "white" theme, the reader should always use the light interface style, even if the device is in dark mode

also, i think the transition pages should also have the same backgrounds colors that the pages have

rbertus2000:
> what i meant was that there shouldn't be a separate option for light and dark themes. it should only be a single theme option
> 
> and then additionally (optionally) an override for the theme appearance
> 
> * if this exists then the "white" and "black" themes would not need to exist, and system could just be "default". overriding the theme appearance should also override the reader controller interface style
> * if this doesn't exist, the themes should have an overridden interface style. e.g. if you select the "white" theme, the reader should always use the light interface style, even if the device is in dark mode
> 
> also, i think the transition pages should also have the same backgrounds colors that the pages have

So instead of the dark mode setting you want an "Override dark mode" toggle that keeps the (light) mode settings active in dark mode?

rbertus2000:
Changes like discussed in Discord.

Aidoku/Features/Reader/Readers/Text/ReaderTextTheme.swift
```diff
@@ -0,0 +1,90 @@
...
+            case "light": .light
+            case "dark": .dark
+            default: .unspecified
+        }
+    }
+
+    /// The current theme's colors, pre-resolved against the appearance override
+    /// (trait propagation doesn't reliably re-resolve dynamic colors in the reader).
+    static var background: UIColor { resolve(current.backgroundColor) }
+    static var text: UIColor { resolve(current.textColor) }
```
Skittyblock: rather than these being computed properties, it would be clearer to have them be functions. properties ideally shouldn't change depending on external state (e.g. userdefaults in this case)

e.g. do `ReaderTextTheme.getCurrentBackground()` rather than `ReaderTextTheme.background`

Aidoku/Features/Reader/ReaderViewController.swift
```diff
@@ -781,6 +790,69 @@ extension ReaderViewController {
...
+        navigationController?.overrideUserInterfaceStyle = styleOverride
+        // presented sheets (e.g. reader settings) don't inherit the override
+        presentedViewController?.overrideUserInterfaceStyle = styleOverride
+        let themed = isTextReader && (theme != .default || styleOverride != .unspecified)
+
+        // the bar renders from the appearance objects set in configure(), which
+        // don't follow the trait override (they resolve against the device style),
+        // so write the theme colors (pre-resolved against the appearance override)
+        // into them directly: the bar background matches the page (like the books
+        // app) and the title uses the theme text color
```
Skittyblock: this is too much text that i think is confusing

my thoughts about comments: https://discord.com/channels/927984899472883723/950039656857223269/1541361659120189460

Aidoku/Features/Reader/Readers/Text/Paged/ReaderPagedTextViewController.swift
```diff
@@ -110,9 +111,13 @@ class ReaderPagedTextViewController: BaseObservingViewController {
         let textSettingChanged: (Notification) -> Void = { [weak self] _ in
             self?.updateTextConfig()
         }
-        for key in ["Reader.textFontSize", "Reader.textLineSpacing", "Reader.textHorizontalPadding", "Reader.textFontFamily"] {
+        for key in [
+            "Reader.textFontSize", "Reader.textLineSpacing", "Reader.textHorizontalPadding",
+            "Reader.textFontFamily", ReaderTextTheme.changeNotification
+        ] {
```
Skittyblock, line 119: should just be one per line to match code style

Aidoku/Features/Reader/Readers/Text/Paged/ReaderPagedTextViewController.swift
```diff
@@ -110,9 +111,13 @@ class ReaderPagedTextViewController: BaseObservingViewController {
...
             addObserver(forName: key, using: textSettingChanged)
         }
+
```
Skittyblock: ?
```suggestion
```

Skittyblock (changes requested):
looks good overall, just small comments

rbertus2000:
> looks good overall, just small comments

done

Skittyblock (approved):

Issue #934, opened 2026-03-16
Text reader issues


Skittyblock:
The text reader is still marked as experimental due to some remaining issues. I think if the following things are fixed it can be considered no longer experimental:

- [ ] The text/content shouldn't shift up and down in the scroll reader
- [ ] Tap zones support should be added
- [ ] Keyboard navigation should also include transition pages, not just transition immediately
- [ ] Trying to transition from a text chapter to non-text chapter should be disabled completely (in the paged reader it has a scroll back to the current chapter, but it should just disable swiping as if there is no text page)
- [ ] Read position saving and restoring should be better when switching between scroll and text reader styles

There may be some other things that I'm missing.

maerlin:
Hi! I'd like to help close out the remaining items here so the text reader can drop the experimental label. My plan, in order:

1. **A small stability PR first (independent of the checklist)**: Fix a force-unwrap crash in `TextPaginator.adjustRangeToBreakPoint` (`UnicodeScalar(nextChar)!` traps on surrogate pairs, e.g. any emoji in an epub). Guard a potential empty-pagination index crash in `ReaderPagedTextViewController.repaginate`. Replace the `.systemRed` debug fallback page. Use the tolerant `EpubParser.entry(in:path:)` lookup in the four reader sites that currently use exact-match `archive[filePath]`, since image-only epub chapters render blank when entry paths are `./`-prefixed or percent-encoded. Plus first unit tests for `TextPaginator`.
2. **Keyboard navigation on transition pages**: `moveLeft()` and `moveRight()` navigate relative to a stale `currentPageIndex` when a transition page is showing. I'll make them transition-page-aware.
3. **Disabling text ↔ non-text chapter transitions**: I'd preload the adjacent chapter's pages when a chapter loads. When the neighbor is non-text, I'd return `nil` from the page view controller data source so the swipe just rubber-bands with no transition page at all. Question: do you prefer full suppression (no transition page), or keeping the info page but making it terminal (can't swipe past it)? "Disabled completely" reads like the former to me, but wanted to check before building it.
4. **Seamless transitions in the paged text reader**: Building on the preload from step 3, this will pre-paginate adjacent text chapters with the same geometry/config. It will serve real first/last-page previews past the info pages (like the paged image reader) and swap chapter state when the user lands on a preview instead of reloading.

Happy to adjust scope or split differently if you'd rather see it another way.

maerlin:
#1033 and #1035 cover the first two items.

[...]

PR #1031, opened 2026-07-24
feat(epub): add epub reading functionality to komga and kavita

### LINKED ISSUE #1022: Komga/Kavita EPUB support

rbertus2000:
Closes #1022

## Summary

Adds EPUB reading support to the Komga and Kavita sources. Previously, EPUB books were filtered out of Komga chapter lists (unless Divina-compatible) and skipped entirely for Kavita. They now show up as regular chapters and open in the text reader, matching the experience of the local files source.

## How it works

EPUBs are not downloaded. Instead, both servers' book APIs are used to split each book into its logical chapters and fetch content on demand:

- **Komga**: The WebPub manifest (`api/v1/books/{id}/manifest`) provides the spine and TOC, which are grouped into chapters the same way the local `EpubParser` does (spine files with a TOC title start a new chapter). Chapter content is fetched from the `resource/` endpoint as XHTML and converted to markdown. Chapter keys use the format `<bookId>/<content file path>`.
- **Kavita**: The book TOC (`api/book/{id}/chapters`) maps spine page indices to titles; content comes from `book-page?page=N`. Chapter keys use the format `<chapterId>/<startPage>-<endPage>`.

Rendering reuses the existing local EPUB pipeline: `EpubParser.segments(fromHTML:)` converts XHTML to text/image segments, and a new shared `EpubParser.pages(...)` builds the reader pages for both local and server sources (the data source only supplies image lookup closures). Text chapters become a single markdown text page with inline images (downloaded to the existing `EpubImages` cache, since the text readers only render local file URLs); image-only chapters (covers, illustration galleries) become regular image pages served through the source's authenticated image requests.

Since books read left-to-right, series containing EPUBs default to LTR — for Komga only when the server doesn't specify a reading direction.

## Also fixed

- `KomgaHelper` now supports custom `Accept` headers; the manifest endpoint returns `application/webpub+json` and responds with 406 to `application/json`.
- Kavita volumes with fractional numbers (e.g. 9.5) were displayed truncated (9), because only the legacy integer `number` field was decoded. The model now prefers `minNumber`. This affected all Kavita series, not just EPUBs.

## Notes

- No project file changes; all code lives in existing files.
- Existing chapter keys are unaffected: non-EPUB books and Divina-compatible EPUBs keep their current keys and behavior.
- Tested against real Komga and Kavita servers as well as the public demo servers (demo.komga.org, demo.kavitareader.com), including text chapters with inline illustrations, image-only cover chapters, EPUB 2/3 TOC variants, and books without a TOC.
- The second part of #1022 (external sources providing direct EPUB links) is not included — that needs an AidokuRunner API change and would make sense as a follow-up.

### DIFF
--- a/Shared/Sources/Kavita/KavitaHelper.swift
+++ b/Shared/Sources/Kavita/KavitaHelper.swift
@@ -167,6 +167,37 @@ struct KavitaHelper: Sendable {
         }
     }
 
+    // Fetch a raw string response (e.g. epub chapter html).
+    func requestString(path: String) async throws(SourceError) -> String {
+        let url = try getServerUrl(path: path)
+
+        func doRequest() async throws(SourceError) -> String? {
+            var request = URLRequest(url: url)
+            guard authorize(request: &request) else {
+                throw SourceError.message("NOT_LOGGED_IN")
+            }
+            guard
+                let result = try? await URLSession.shared.data(for: request),
+                let response = result.1 as? HTTPURLResponse
+            else {
+                throw SourceError.networkError
+            }
+            if response.statusCode == 401 {
+                return nil
+            }
+            return String(data: result.0, encoding: .utf8)
+        }
+
+        if let result = try await doRequest() {
+            return result
+        }
+        // retry after re-auth
+        guard try await refreshToken(), let result = try await doRequest() else {
+            throw SourceError.message("NOT_LOGGED_IN")
+        }
+        return result
+    }
+
     func refreshToken() async throws(SourceError) -> Bool {
         let url = try getServerUrl(path: "api/account/refresh-token")
 
--- a/Shared/Sources/Kavita/KavitaHelper.swift
+++ b/Shared/Sources/Kavita/KavitaHelper.swift
@@ -61,6 +61,53 @@ struct KavitaHelper: Sendable {
         body: Data? = nil,
         lastWorkingMirror: inout URL?
     ) async throws(SourceError) -> T {
+        let data = try await requestData(path: path, method: method, body: body, lastWorkingMirror: &lastWorkingMirror)
+
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .custom({ decoder in
+            let container = try decoder.singleValueContainer()
+            let string = try container.decode(String.self)
+            let formatter = DateFormatter()
+            formatter.timeZone = if #available(iOS 16.0, macOS 13.0, *) {
+                .gmt
+            } else {
+                .init(secondsFromGMT: 0)
+            }
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSSSSS"
+            var date = formatter.date(from: string)
+            if date == nil {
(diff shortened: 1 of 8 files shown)


Shared/Sources/Kavita/KavitaHelper.swift
```diff
@@ -167,6 +167,37 @@ struct KavitaHelper: Sendable {
         }
     }
 
+    // Fetch a raw string response (e.g. epub chapter html).
+    func requestString(path: String) async throws(SourceError) -> String {
```
Skittyblock: this seems to be largely duplicated code (also doesn't handle mirrors) so i think it would make sense to refactor things a bit to reduce it... maybe make `request` use a request data function which handles mirrors and auth, and then just adds the decoding to the end, and then have the request string just convert the data to a string

Shared/Sources/Kavita/KavitaModels.swift
```diff
@@ -237,35 +237,109 @@ struct KavitaVolume: Codable, Sendable {
...
+        var titles: [Int: String] = [:]
+        func walk(_ items: [KavitaBookChapterItem]) {
+            for item in items {
+                if titles[item.page] == nil, !item.title.isEmpty {
+                    titles[item.page] = item.title
+                }
+                walk(item.children ?? [])
+            }
+        }
+        walk(toc)
```
Skittyblock: i think it's generally better to avoid recursion and just append children to one array/stack you can continue iterating over until empty

Shared/Sources/Kavita/KavitaSource.swift
Skittyblock, line 104: swiftlint warnings should be resolved; you can break the body into multiple extensions

Skittyblock (changes requested):

rbertus2000:
All points addressed. No more linter errors.

Skittyblock:
after considering this more i think splitting epubs up into individual chapters is not really something we should aim to do. i don't think there's a way to always get it to work well for every epub, and just leaving it entirely in a single "chapter"/volume would be a better reflection of komga/kavita anyways. i was okay with the local files source doing it for now as a temporary implementation (since i am not itching to work on it myself), but i worry more about doing it with komga/kavita since rolling it back in the future if it does end up changing would be more of a hassle

maerlin:
Hi, I have been working through the text reader items in #934 (#1033 and #1035) and wanted to flag something rather than risk duplicating your effort.

I got chatting with skitty about epubs and we landed on the idea of a separate reader: its own paged UI, readium-css for styling only, XHTML rendered in a locked down WKWebView served from the zip through a custom scheme handler, keeping the existing `EpubParser`. I spiked it last week and it went well, though it turned up a few non-obvious traps.

Worth naming the overlap plainly: if that lands, `EpubParser.segments(fromHTML:)` and the html to markdown conversion go away, Komga and Kavita included. Nothing of mine is written beyond a throwaway spike, and I have no interest in cutting across work you have already done.

What are your plans for this PR? If it is mostly a matter of time, I would rather help get it over the line than start something that collides with it.

rbertus2000:
Superseeded by #1052

PR #1033, opened 2026-07-30
fix: text reader crash fixes and tolerant epub archive lookup

maerlin:
First step of the text reader stability work discussed in #934 (see my comment https://github.com/Aidoku/Aidoku/issues/934#issuecomment-5131518658): crash and robustness fixes only, no behaviour changes.

## Changes
- Fix a force-unwrap crash in `TextPaginator`'s sentence-break search. The character following a sentence ender is read as a UTF-16 code unit, and `UnicodeScalar(UInt16)` returns nil for surrogate halves, which was force-unwrapped. Any epub containing an emoji near a page break could crash the paged reader.
- Guard against empty pagination results in `ReaderPagedTextViewController.repaginate()` before indexing; the `pages.count - 1` paths could index with -1.
- Replace a `sections.last!` force-unwrap in the scroll reader's `appendNextChapter`.
- Use a neutral background for the paged reader's out-of-range fallback page instead of `.systemRed`.
- Add a tolerant zip entry lookup and use it at the sites that previously read entries with the exact-match subscript, so entries written with a `./` prefix or with percent-encoded paths (common when file names contain spaces or non-ASCII characters) load instead of rendering blank.

## Review feedback
The tolerant lookup was originally added to `EpubParser`. It now lives in `Shared/Extensions/Archive.swift` as `Archive.entry(at:)`, since the behaviour concerns zip entry lookup rather than epub parsing. `EpubParser` calls it through the same extension.

## Testing
- Added the first `TextPaginator` unit tests, including a regression test for the surrogate crash. It crashes the test runner on the previous code and passes with the fix.
- Added `ArchiveTests` covering `entry(at:)`: a `./` prefixed entry, a percent-encoded entry path, case insensitivity, an exact match, and a genuinely absent path. Lookups always arrive decoded, because `EpubParser.resolve(href:relativeTo:)` decodes every href, so the encoded case is a property of the archive rather than of the reference.
- Verified in the simulator with an epub whose entries all carry a `./` prefix. Its image-only chapter renders, which is the path that previously used the exact-match subscript and returned nil.
- Full `AidokuTests` suite passes on the iOS 26.5 simulator, both targets build clean, and there are no new SwiftLint warnings.

### DIFF
--- a/Shared/Extensions/Archive.swift
+++ b/Shared/Extensions/Archive.swift
@@ -1,0 +1,29 @@
+//
+//  Archive.swift
+//  Aidoku
+//
+//  Created by Pietro Baiguini on 8/9/26.
+//
+
+import ZIPFoundation
+
+extension Archive {
+    /// Looks up an entry by path, tolerating the variations archives use for the same file.
+    ///
+    /// The subscript requires an exact match, so entries written with a `./` prefix or with
+    /// percent-encoded characters (common when file names contain spaces or non-ASCII
+    /// characters) fail to resolve. This falls back to a case-insensitive comparison against
+    /// both the normalised and the percent-decoded path.
+    func entry(at path: String) -> Entry? {
+        if let entry = self[path] { return entry }
+        let target = path.lowercased()
+        return first { entry in
+            var entryPath = entry.path
+            if entryPath.hasPrefix("./") {
+                entryPath = String(entryPath.dropFirst(2))
+            }
+            return entryPath.lowercased() == target
+                || (entryPath.removingPercentEncoding ?? entryPath).lowercased() == target
+        }
+    }
+}
--- a/iOS/UI/Reader/Page/ReaderPageView.swift
+++ b/iOS/UI/Reader/Page/ReaderPageView.swift
@@ -357,7 +357,7 @@ class ReaderPageView: UIView {
             do {
                 var data = Data()
                 let archive = try Archive(url: zipURL, accessMode: .read)
-                guard let entry = archive[filePath] else {
+                guard let entry = EpubParser.entry(in: archive, path: filePath) else {
                     return nil
                 }
                 _ = try archive.extract(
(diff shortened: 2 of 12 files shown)


iOS/UI/Reader/Page/ReaderPageView.swift
```diff
@@ -357,7 +357,7 @@ class ReaderPageView: UIView {
             do {
                 var data = Data()
                 let archive = try Archive(url: zipURL, accessMode: .read)
-                guard let entry = archive[filePath] else {
+                guard let entry = EpubParser.entry(in: archive, path: filePath) else {
```
Skittyblock: this should be refactored into probably an extension on ZIPFoundation's `Archive` instead of on `EpubParser` if it's reused, since this is not something related to epub parsing

Skittyblock (changes requested):
thank you, these are things i missed when reviewing the other text reader prs. looks good, i would just recommend moving the `EpubParser.entry` function

iOS/UI/Reader/Page/ReaderPageView.swift
maerlin: Moved in e372dbf5. It is now `Archive.entry(at:)` in `Shared/Extensions/Archive.swift`. `EpubParser` and `LocalFileManager`, which both used the previous helper, call it through the extension.

Added `ArchiveTests` in b529cc0c covering the `./` prefix and percent-encoded entry cases.

Shared/Extensions/Archive.swift
```diff
@@ -0,0 +1,29 @@
...
+    func entry(at path: String) -> Entry? {
+        if let entry = self[path] { return entry }
+        let target = path.lowercased()
+        return first { entry in
+            var entryPath = entry.path
+            if entryPath.hasPrefix("./") {
+                entryPath = String(entryPath.dropFirst(2))
+            }
+            return entryPath.lowercased() == target
+                || (entryPath.removingPercentEncoding ?? entryPath).lowercased() == target
```
Skittyblock, line 26: could avoid a duplicate check you already know would fail
```suggestion
                || entryPath.removingPercentEncoding.flatMap { $0.lowercased() == target } ?? false
```

Skittyblock (approved):
ty

PR #1035, opened 2026-07-31
fix: correct keyboard navigation on text reader transition pages

maerlin:
Second step of the text reader work from #934: this covers the "keyboard navigation on transition pages doesn't work correctly" item.

## The problem
Chapter transition pages are not part of the paginated `pages` array, so while one is displayed `currentPageIndex` still refers to the page the reader came from. `moveLeft()` and `moveRight()` computed their target by stepping from that stale index, which landed one page off:
- On the next-chapter transition page, moving backward went to the second-to-last page rather than the last one (the case described in the issue).
- On the previous-chapter transition page, moving forward skipped to the second page rather than the first.

Moving toward the adjacent chapter happened to work already, but only because the out-of-range index fell through to a transition helper which then detected it was already on the transition page.

## The fix
Following your review: `moveLeft` and `moveRight` now ask the data source for the adjacent view controller, in the same way `ReaderPagedViewController` does, rather than computing a target index. The data source already resolves every transition case correctly:

- `viewControllerBefore` a next-chapter transition page returns the last text page.
- `viewControllerAfter` a previous-chapter transition page returns the first text page.
- Moving toward the adjacent chapter returns the load trigger page, and the existing `didFinishAnimating` handling performs the transition.
- Either side of the load trigger page returns nil, so input is ignored while a chapter load is in flight.

Because the target now comes from the data source, `navigateToPreviousChapterTransition` and `navigateToNextChapterTransition` have no remaining callers and are removed. They duplicated the transition page construction the data source already performs. The `move` function keeps its original signature; the added `direction` parameter is gone, since nothing computes a target index any more. Keyboard input now takes the same path as a swipe.

Net change against main is 34 lines added and 86 removed, in one file.

Tap zones and the Apple Pencil squeeze gesture route through the same two methods, so they are covered as well.

## Two things left alone
Both are pre-existing and outside this fix, so I have not touched them. Happy to address either separately if you want them.

- `effectiveReadingMode` is hardcoded to `.ltr`, so every `case .rtl` branch in this file is unreachable.
- The text reader has no equivalent of the paged reader's `isTransitioning` re-entrance guard. The data source returning nil for the load trigger page covers the in-flight chapter load this fix concerns, but rapid input during an animation is still unguarded.

## Testing
Verified manually in the paged text reader on the iOS 26.5 simulator, against a multi-chapter epub in the local files source:

- Moving backward off the next-chapter transition page lands on the last page of the chapter, and moving forward off the previous-chapter transition page lands on the first. These are the two cases that were off by one.
- Moving toward the adjacent chapter from either transition page still loads it.
- Input during a chapter load is ignored, with no queued moves and no double advance.
- At the first and last chapters the transition page is shown, and further input in the same direction does nothing.
- Tap zones behave identically to the arrow keys, since both route through `moveLeft` and `moveRight`.
- The animation direction when leaving a transition page is correct in both directions.
- Ordinary page stepping, the slider, and double page mode are unaffected.

Both targets build clean, the full `AidokuTests` suite passes, and there are no new SwiftLint warnings.

Independent of #1033. The two touch different regions of the file and can be merged in either order.

### DIFF
--- a/iOS/UI/Reader/Readers/Text/Paged/ReaderPagedTextViewController.swift
+++ b/iOS/UI/Reader/Readers/Text/Paged/ReaderPagedTextViewController.swift
@@ -357,7 +357,7 @@ class ReaderPagedTextViewController: BaseObservingViewController {
 
     // MARK: - Navigation
 
-    func move(toPage index: Int, animated: Bool) {
+    func move(toPage index: Int, animated: Bool, direction explicitDirection: UIPageViewController.NavigationDirection? = nil) {
         guard !pages.isEmpty else {
             return
         }
@@ -374,8 +374,12 @@ class ReaderPagedTextViewController: BaseObservingViewController {
 
         let viewController = createPageViewController(for: targetIndex)
 
+        // Moving off a transition page can target the page we're already on,
+        // so the caller supplies the direction in that case
         let direction: UIPageViewController.NavigationDirection
-        if effectiveReadingMode == .rtl {
+        if let explicitDirection {
+            direction = explicitDirection
+        } else if effectiveReadingMode == .rtl {
             direction = targetIndex < oldIndex ? .forward : .reverse
         } else {
             direction = targetIndex > oldIndex ? .forward : .reverse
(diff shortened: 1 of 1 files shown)


Skittyblock (commented):
the logic here seems to be getting a little complicated... couldn't we just utilize the `viewControllerBefore` and `viewControllerAfter` functions in the same way that the paged reader does?
https://github.com/Aidoku/Aidoku/blob/dc7164d9e9764606a227e4d3dce1c2bb1fc1da16/iOS/UI/Reader/Readers/Paged/ReaderPagedViewController.swift#L770-L791
this would reduce the code duplication. i'm not sure why the `move` function is even necessary given that it seems to perform similar computation that the `pageViewController` `didFinishAnimating` does

maerlin:
Rewritten in ed5da6ef. `moveLeft` and `moveRight` now ask the data source for the adjacent view controller, as `ReaderPagedViewController` does, rather than computing an index.

The data source already resolved every transition case, so `navigateToPreviousChapterTransition` and `navigateToNextChapterTransition` had no remaining callers and are removed, along with the `direction` parameter I had added to `move`. Keyboard input now takes the same path as a swipe. Net change against main is 34 lines added and 86 removed.

Verified in the simulator: moving off either transition page lands on the correct page, transitions still load in both directions, and input during a load is ignored. Two pre-existing points are noted in the description.

Skittyblock (approved):

PR #1040, opened 2026-08-08
fix: selected cells appear unselected (and reverse)

### LINKED ISSUE #977: Selected cells sometimes appear visually unselected (and vice versa)
this is some bug with swiftui that i unfortunately don't know how to fix

one way to reproduce (from series view):
- turn chapter selection on
- press select all
- scroll down
- press deselect all
- scroll up
- chapters will look selected even though they aren't

another (from source migrate view):
- select source with multiple entries to migrate
- press continue without changing series selction
- press the back button
- series will appear all unselected despite being selected

affects both chapter selection and migration series selection

could be fixed by rolling our own selection ui, but i don't want to lose the multi select gestures and dragging along the checkboxes to select

Amqx:
fixes #977 

this appears to be an already identified swift ui bug - see the following: [1](https://developer.apple.com/forums/thread/677802), [2](https://developer.apple.com/forums/thread/682879), and [3](https://developer.apple.com/forums/thread/707333).

from my own testing, this is what i think is happening:
for reproduction 1: when the deselect all call happens, swiftui lazily deselects by only telling uikit to turn off what is visible. when you scroll up, it conveniently forgets to tell uikit to deselect previously unrendered indices.
for reproduction 2: the new view controller detaches the list's host view from the window, then uikit silently drops everything it knows about what was selected. afterwards, when swiftui sends the list back, since the list never changed, it doesn't request a re-render, causing uikit's empty state to get realized.

i have provided 2 bandaids, of which neither are exactly ideal:

- for the series view (unselected looks selected) we manually drain everything out of uikit's memory
- for the migration view (selected looks unselected) we remove then immediately re-add an entry to force a re-render

### DIFF
--- a/iOS/New/Utilities/Modifiers/RestoreListSelectionOnAppear.swift
+++ b/iOS/New/Utilities/Modifiers/RestoreListSelectionOnAppear.swift
@@ -1,0 +1,37 @@
+//
+//  RestoreListSelectionOnAppear.swift
+//  Aidoku
+//
+//  Created by Amqx on 8/6/26.
+//
+
+import SwiftUI
+
+// Remove then re-add an entry to force a re-render.
+private struct RestoreListSelectionOnAppearModifier<Value: Hashable>: ViewModifier {
+    @Binding var selection: Set<Value>
+
+    @State private var appeared = false
+
+    func body(content: Content) -> some View {
+        content
+            .onAppear {
+                // the first appearance renders from scratch, so there's nothing to restore
+                guard appeared else {
+                    appeared = true
+                    return
+                }
+                guard let value = selection.first else { return }
+                selection.remove(value)
+                DispatchQueue.main.async {
+                    selection.insert(value)
+                }
+            }
+    }
+}
+
+extension View {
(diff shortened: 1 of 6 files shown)


iOS/New/Utilities/Modifiers/RestoreListSelectionOnAppear.swift
```diff
@@ -0,0 +1,60 @@
...
+                guard appeared else {
+                    appeared = true
+                    return
+                }
+                guard let value = selection.first else { return }
+                // with a single entry selected the selection is briefly empty, which would otherwise
+                // look like a deselection and wipe the selection in the backing collection view
+                restoreState.isRestoring = true
+                selection.remove(value)
+                DispatchQueue.main.async {
```
Skittyblock: swift concurrency should be preferred over gcd

could also make this a `.task` instead of `.onAppear` and do `await MainActor.run`, but i don't believe there's any notable difference
```suggestion
                Task { @MainActor in
```

Skittyblock (changes requested):
the `restoresListSelectionOnAppear` is a little buggy with the animation, but a solution for this is much appreciated, thank you

Skittyblock (approved):

PR #1042, opened 2026-08-08
perf: add core data fetch indexes for library and history queries

Amqx:
currently there are absolutely 0 indexes for data, so when the app starts up it does a ton of full row scans for each entry. this adds indexes which massively increases the time data starts showing on the library and history tabs (~4s for entries to start showing vs pretty much instant, tested on a 251 library entries + 65k history entries + 130k history entries backup)


Amqx:
Note this also massively speeds up a lot of other stuff

As an example, marking lots of chapters as read at the same time.

Every time this looks up a chapter to mark history it does a full table scan for every entry so it takes forever.

Ex. Go to series with 500+ chapters, then attempt to mark all as read. Before this takes idk how long cause I didn't bother waiting for it to finish after a minute, after the indexes patch it takes like 4 seconds. 

There's another bug for this where marking read takes more operations than required but I'll fix that in another pr.

Shared/Managers/CoreData/CoreDataManager.swift
```diff
@@ -82,6 +82,11 @@ final class CoreDataManager {
         container.viewContext.automaticallyMergesChangesFromParent = true
         container.viewContext.mergePolicy = NSMergePolicy(merge: .mergeByPropertyObjectTrumpMergePolicyType)
 
+        // stores that predate the model's fetch indexes don't migrate to pick them up
+        if let cloudStoreUrl = cloudDescription.url {
+            Self.createFetchIndexes(at: cloudStoreUrl)
+        }
```
Skittyblock: really? they don't get added during automatic migrations? that's surprising to me, but i guess it's not mentioned here https://developer.apple.com/documentation/coredata/migrating-your-data-model-automatically

Skittyblock (commented):
wow thanks, i didn't know this was a thing

the next planned version is actually 0.9, and i've added a new schema in the ocr dictionary pr which i should  *hopefully*  merge soon, so this will have to be adjusted to add to the 0.9 schema instead of a new 0.8.5 one after that happens

Shared/Managers/CoreData/CoreDataManager.swift
Amqx: really isn't since fetch indexes aren't part of the model's version hash, so since I only added indexes it isn't a schema bump as far as core data thinks. you can test with momc:

```bash
# from repo root
MOMC=$(xcrun --find momc)
SDK=$(xcrun --sdk macosx --show-sdk-path)
"$MOMC" --sdkroot "$SDK" --module Aidoku Shared/Aidoku.xcdatamodeld output.momd
# gives the following output:
.....
0.8.3.xcdatamodel: note: Model 0.8.3 version checksum: vawzC42JY3VPhHRsOIXloBc0E7PBZWcTcriCC8s510w=
.....
0.8.5.xcdatamodel: note: Model 0.8.5 version checksum: vawzC42JY3VPhHRsOIXloBc0E7PBZWcTcriCC8s510w=
.....
```

as you can see the checksums come out the same, so `isConfiguration(withName:, compatibleWithStoreMetadata:)` returns true for existing stores, so no migration happens, meaning core data never creates the new schema and indexes never appear.

however, core data DOES migrate if the schema comes with an actual change that isn't indexes. since you intend to add these changes to 0.9.0 the migration will run, so you should be able to drop these if you'd like.

PR #1043, opened 2026-08-14
fix: reschedule library refresh after background task completes

Zareix:
It looks like the refresh job might not be rescheduling properly, which could be causing library refresh and notifications to break after just one job.
This seems to give more consistent results with notifications.


Skittyblock (approved):
lgtm, thanks

PR #1044, opened 2026-08-15
feat: continue reading when reselecting history tab

Amqx:
when the history tab is reselected the following behaviour now occurs

1. (existing) if not in full history list (ie inside a specific manga) return to history list
2. (existing) if not at top of history list return to top of history list
3. (new) if at top of full history list, resume reading last read series

this is guarded by a setting in library that defaults to false, so users have to opt-in

### DIFF
--- a/Shared/Localization/en.lproj/Localizable.strings
+++ b/Shared/Localization/en.lproj/Localizable.strings
@@ -461,6 +461,7 @@
 "READER" = "Reader";
 "OPEN_READER_VIEW" = "Open Reader View";
 "RESUME_LAST_OPENED_CHAPTER" = "Resume Last Opened Chapter";
+"CONTINUE_READING_ON_RESELECT" = "History Doubleclick Resumes";
 "UNREAD_CHAPTER_BADGES" = "Unread Chapter Badges";
 "DOWNLOADED_CHAPTER_BADGES" = "Downloaded Chapter Badges";
 "PIN_TITLES" = "Pin Titles";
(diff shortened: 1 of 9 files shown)


iOS/New/Views/History/HistoryView.swift
```diff
@@ -247,6 +255,74 @@ struct HistoryView: View {
         }
     }
 
+    // open the most recently read manga, resuming from the chapter it left off on
+    @MainActor
```
Skittyblock: `View` functions are `MainActor` isolated by default so you don't need to mark it explicitly
```suggestion
```

Skittyblock (changes requested):
i don't think this needs to be hidden behind a setting that's off by default. it seems like natural enough ux that it shouldn't cause any issues for people

iOS/New/Views/History/HistoryView.swift
Amqx: applied

Skittyblock (changes requested):
oh i meant that i don't think it needs to be a setting at all, it can just be the standard behavior

Amqx:
i feel like leaving it as a setting is better
1. it's not the most obvious behaviour, so the setting allows some users to actually discover it
2. i guarantee you someone is going to not like it and want to turn it off cause they keep fat fingering it

reason 2 was why i was split 50/50 on leaving it default off

Shared/Localization/en.lproj/Localizable.strings
```diff
@@ -461,6 +461,7 @@
 "READER" = "Reader";
 "OPEN_READER_VIEW" = "Open Reader View";
 "RESUME_LAST_OPENED_CHAPTER" = "Resume Last Opened Chapter";
+"CONTINUE_READING_ON_RESELECT" = "History Doubleclick Resumes";
```
Skittyblock: in that case this might be a bit better name, since 'doubleclick' isn't a word
```suggestion
"CONTINUE_READING_ON_RESELECT" = "History Tab Double Tap Resumes";
```

PR #1045, opened 2026-08-15
feat: add partial backup decoding

Amqx:
currently every time the app needs the details of a backup it does a full decode. for larger backups (>20MB), this is very slow (maybe 1-2s/ backup), and you must wait for any backups above your intended one to load before accessing what you want.

this adds a partial decoder specifically for the plist backups aidoku makes. we decode the binary blob at the tail of the plist to get an offset table, then manually decode exactly what we want.

for counts (library entries, history, chapters, etc.), the headers for each container object contain their element counts, so we can take them directly without parsing the container.

for name/ date, we just parse them directly.

if it's a json/xml backup this falls back to the original full load since those don't contain counters.


Skittyblock (approved):
didn't know this was doable with binary plists, thanks. haven't tested but lgtm

PR #1050, opened 2026-08-20
fix: keep incomplete downloads in a failed state

### LINKED ISSUE #243: Missing pages when download interrupted and unable to tell
I noticed some of the downloaded chapters are missing pages even when the status is complete. To fix it you need to remove the download and redownload. Just wondering whether there is a better way to check whether all downloads are complete

maerlin:
Closes #243.

A chapter that finished downloading with pages missing was promoted exactly like a complete one, so it showed as downloaded and the gap was found only on reading it. This keeps such a download in a failed state instead, and represents that state in the manga details view and in the downloads settings page.

## Changes
- Make `handleChapterDownloadFinish` a three-way gate. It previously discarded a download only when every page failed; any other failure count took the branch that logs to `LogManager` and then promotes the staging directory. A download with some pages missing now keeps its staging directory, marks it, and reports `.failed` to the delegate.
- Mark a failed download with a `.failed` sentinel file inside its `.tmp_` directory, holding the numbers of the pages that failed. The marker sits inside the directory so that discarding the download remains a single `removeItem`, and is named so that it cannot be mistaken for a page.
- Return `.failed` from `getDownloadStatus` for a marked staging directory and `.queued` for an unmarked one. `DownloadStatus.failed` already existed in the enum and was never produced; a failed partial and a download in progress were identical on disk and both reported `.queued`.
- Add `DownloadTaskDelegate.downloadFailed` and a `.downloadFailed` notification, observed by the manga view model and by both downloads settings view models.
- Track which pages failed rather than how many. `NetworkPage` carries its one-based page number and `downloadPage` returns it with the result, since pages are fetched concurrently and complete out of order.
- Show a failed chapter in the manga details view with its own indicator. It stays locked, keeps its download button for retry, and offers removal.
- List a failed download in the downloads settings page with an "Incomplete Download" subtitle, and count its size, so the space it occupies can be seen and reclaimed. Its row is not tappable, because opening it would fall through to the source and read the chapter online. Chapter metadata is written into the staging directory on failure so that the page has a title to show.
- Discard whatever an earlier attempt left behind when a failed chapter is queued again. A retry refetches every page and the stored extension comes from the response, so keeping the earlier files risks a page arriving as a jpeg where it was a png before, leaving the promoted chapter holding that page twice.
- Remove a marked staging directory in `delete(chapters:)`, which is the only way back out of the failed state.

## Hidden staging directories
`URL.contents` passes `.skipsHiddenFiles`, and a staging directory is named `.tmp_<key>`. No listing has ever reported one, so every filter written against that prefix had nothing to match. A new `contentsIncludingHidden` is used where a failed download has to be visible. Two consequences beyond the new UI:

- `hasRemainingChapters` in `delete(chapters:)` counted only visible entries, so removing the last visible chapter of a manga concluded that the manga was empty, deleted the manga directory, and took any staging directory with it.
- The sweeps that clear staging directories when downloads are cancelled were dead for the same reason. They now run, with marked directories excluded so that cancelling a download does not delete a chapter that failed earlier. This is a behaviour change beyond the issue itself; I can move it to a separate PR if you prefer, though leaving dead code directly under a guard added here seemed worse.

## Out of scope
- Resume. A retry refetches every page. The marker already records which pages are missing and page files are named deterministically, so refetching only those is a small follow-up.
- A download interrupted by the app being killed leaves an unmarked staging directory, which still reports `.queued` indefinitely. This is pre-existing and untouched here.

## Testing
- Added `PartialDownloadTests` covering the status gate (none, queued for an unmarked staging directory, failed for a marked one, finished for a chapter directory), the marker's position inside the staging directory, and the discard behaviour on retry.
- Verified in the simulator against the Komga demo server. A chapter cut off mid-download ends failed, and the marker names exactly the pages missing from disk.
- Both recovery paths produce a correct archive, with contiguous pages and `ComicInfo.xml`, no marker inside the `.cbz`, and no duplicates from the reused staging directory. This covers retrying in place, and removing then downloading again.
- Several failed chapters selected together offer Download rather than Remove, and retrying clears all of them.
- Removing the last complete chapter of a manga leaves a failed partial intact, and Remove All Downloads stays reachable for a manga holding nothing else.
- Full `AidokuTests` suite passes on the iOS 26.5 simulator (365 tests), and there are no new SwiftLint warnings.

### DIFF
--- a/AidokuTests/PartialDownloadTests.swift
+++ b/AidokuTests/PartialDownloadTests.swift
@@ -54,39 +54,6 @@ struct PartialDownloadTests {
         #expect(DownloadManager.shared.getDownloadStatus(for: chapter) == .finished)
     }
 
-    /// A retry refetches every page and the extension comes from the response, so anything the
-    /// failed attempt left behind has to go or the chapter can end up holding a page twice.
-    @Test func discardingAFailedDownloadTakesItsPagesWithIt() throws {
-        let cache = DownloadCache()
-        let chapter = identifier("discarded")
-        defer { removeDirectories(for: chapter) }
-
-        let tmpDirectory = cache.tmpDirectory(for: chapter)
-        tmpDirectory.createDirectory()
-        try Data().write(to: tmpDirectory.appendingPathComponent("001.png"))
-        try Data("[2]".utf8).write(to: cache.failureMarker(inTmpDirectory: tmpDirectory))
-
-        cache.discardFailedDownload(for: chapter)
-
-        #expect(!tmpDirectory.exists)
-    }
-
-    /// Only a failed download is discarded: the same directory shape belongs to a download that is
-    /// still running, and removing that would cancel it.
-    @Test func discardingLeavesAnUnmarkedStagingDirectoryAlone() throws {
-        let cache = DownloadCache()
-        let chapter = identifier("running")
-        defer { removeDirectories(for: chapter) }
-
-        let tmpDirectory = cache.tmpDirectory(for: chapter)
-        tmpDirectory.createDirectory()
-        try Data().write(to: tmpDirectory.appendingPathComponent("001.png"))
-
-        cache.discardFailedDownload(for: chapter)
-
-        #expect(tmpDirectory.exists)
-    }
-
     /// The marker lives inside the staging directory, so discarding a failed download stays a
     /// single removal and nothing is left beside it.
     @Test func markerLivesInsideTheStagingDirectory() {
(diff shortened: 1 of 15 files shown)


Skittyblock (approved):
changes i made for reference:
- remove most comments (generally excess comments makes the code less readable imo)
- filter once instead of twice when possible
- avoid linear scan to check for failed download in `getDownloadedChapters`
- remove `discardFailedDownload` (every time it was being used, the tmp directory was already being fetched so it could just be removed before creating. the extra checks for the failed marker were unnecessary)

PR #1052, opened 2026-08-25
feat: add epub reader

maerlin:
Adds an ePub reader for local files and books served by Komga and Kavita. Right-to-left publications, bookmarks and richer position formats are out of scope.

## Changes

- `ReaderEpubViewController` hosts a whole book rather than one document, so page turns continue across spine boundaries and the toolbar describes the book.
- `ReaderEpubViewModel` owns the reading state, so cross-spine navigation is testable without a view.
- `EpubSpineRenderer` renders one spine document, reusing a single `WKWebView` for the whole spine.
- `EpubSpineMeasurer` measures the rest of the spine in the background, so the book total firms up while reading.
- `EpubPageIndex` keeps the counts as prefix sums, returning `nil` rather than approximating while they're incomplete.
- `EpubTableOfContents` and `ReaderEpubContentsView` present the book's own contents alongside `ReaderChapterListView`.
- `ReaderTableOfContentsReader` refines `ReaderReaderDelegate`, leaving other readers unchanged.
- `EpubSchemeHandler` serves resources over a custom scheme, so nothing is extracted to disk and the web view never gets a file origin.
- `EpubWebViewFactory` disables page JavaScript, uses a non-persistent data store, and blocks all http and https loads.
- `EpubPaginationSettings` maps the existing text reader settings onto readium-css variables.
- `EpubChapterCache` caches server-side books, checking staleness by `Last-Modified` and size, evicting at 500 MB.
- `KomgaSourceRunner` and `KavitaSourceRunner` route ePub chapters to the reader, sharing `KomgaHelper.baseUrls(lastWorkingMirror:)` so downloads walk the same mirrors as requests.
- Clearing the network cache counts and removes downloaded ePubs.

## Notes

- `--USER__noOverflow` is set deliberately. It reads as a double negative: setting it removes readium-css's overflow clipping, so a wide table bleeds rather than being cut off at the column edge. The injection script then scales the table to fit. Upstream: readium/readium-css#138.
- Position is stored as a fraction rather than a pointer into the text. A page index doesn't survive relayout (18pt to 20pt took a 456 page book to 520), and exact restore needs a locator grammar, which is a feature of its own. Both text readers already store a fraction in `HistoryObject.scrollPosition`.

## Testing

- Full suite green on the iPhone 17 simulator, 499 cases, no new SwiftLint warnings.
- Manual pass over the seven-book corpus: paging, rotation both ways, contents, internal links and the return offer, image and table previews, both modes and switching between them, resuming a part-read book.
- Komga and Kavita exercised by hand against their public demo servers.


Aidoku.xcodeproj/project.pbxproj
```diff
@@ -48,7 +48,9 @@
 /* End PBXFileSystemSynchronizedBuildFileExceptionSet section */
 
 /* Begin PBXFileSystemSynchronizedRootGroup section */
-		FB77CD38303A838B00DF3DA3 /* Aidoku */ = {isa = PBXFileSystemSynchronizedRootGroup; exceptions = (FB77CE44303A838C00DF3DA3 /* PBXFileSystemSynchronizedBuildFileExceptionSet */, ); explicitFileTypes = {}; explicitFolders = (); path = Aidoku; sourceTree = "<group>"; };
+		FB77CD38303A838B00DF3DA3 /* Aidoku */ = {isa = PBXFileSystemSynchronizedRootGroup; exceptions = (FB77CE44303A838C00DF3DA3 /* PBXFileSystemSynchronizedBuildFileExceptionSet */, ); explicitFileTypes = {}; explicitFolders = (
+				"App/Resources/readium-css",
```
Skittyblock: i think this is an xcodeproj group instead of a folder? it should just be a regular folder, and i think it should probably be put in the features/readers/epub folder

Aidoku/Features/Reader/Readers/Epub/ReaderEpubViewController.swift
```diff
@@ -0,0 +1,885 @@
...
+
+    private struct WebViewInsets {
+        let top: NSLayoutConstraint
+        let bottom: NSLayoutConstraint
+        let leading: NSLayoutConstraint
+        let trailing: NSLayoutConstraint
+
+        var constants: UIEdgeInsets {
+            UIEdgeInsets(top: top.constant, left: leading.constant,
+                         bottom: bottom.constant, right: trailing.constant)
```
Skittyblock: ideally shouldn't have code that uses alignment like this (my personal preference for aidoku's code style)

```suggestion
            UIEdgeInsets(
                top: top.constant,
                left: leading.constant,
                bottom: bottom.constant,
                right: trailing.constant
            )
```

Skittyblock (changes requested):
haven't looked at the code extensively yet but a few comments:
- there are two swiftlint errors stopping compilation that can be fixed with `// swiftlint:disable:next empty_count` (i think this is a swiftlint bug, i'm not sure how you compiled without it erroring for you. maybe it's just on my end?)
- > no UIPageViewController

  ...is this not possible? the books app uses this and i think eventually we will have to support it (e.g. bc it would be nice to have page turn animations like the page flip as an option in the future)
  - the gestures for flipping pages feel unnatural not following your finger
- the transition between chapters/spines on the scroll style seems a little rough. it looks like the books app does something like the webtoon infinite scroll where it just appends to the end (though it also seems to have caching that makes it instant even if you close the app and open again. not quite sure how it works), which might be worth trying to copy
  - ideally the text shouldn't be cut off in the safe areas, it should look more like the text reader does currently
  - i think the scroll bar should be hidden
  - i don't think the left/right swipe gestures should be enabled in the scroll style. it's better to e.g. allow dismissing the view by swiping off the sides

maerlin:
Fixed:
- swiftlint disables at `ReaderEpubViewController.swift:531` and `ReaderEpubViewModel.swift:66`
- `UIEdgeInsets` formatting
- scroll indicators hidden
- `readium-css` moved to `Features/Reader/Readers/Epub/`

It compiled here because the app target doesn't lint through the plugin. The plugin only covers `AidokuRunner` and `Wasm3`; the app target's run script shells out to a Homebrew swiftlint and skips when there isn't one. I don't have one.

Also: pinned 0.63.3 doesn't flag `count > 0`, and calls all three disables superfluous, including the existing one at `BinaryPlistScanner.swift:196`. Renaming the bindings satisfies both versions if you'd rather drop the comments.

`explicitFolders` is the other way round: it's what makes it a folder instead of a group. Without it the directory flattens into the bundle root and the `subdirectory:` lookup fails. Moved it regardless.

Confirmed and taking next: swipes jump, replacing with a pan on `contentOffset`; side swipes off in scroll style; scroll style text is cut because the safe area goes to the frame rather than `contentInset`; spine transitions appended like the webtoon reader.

`UIPageViewController`: pages are scroll offsets in one web view, so it needs a web view per page or snapshots, and snapshots lose text selection and dictionary lookup. Page flip does need it. This PR or a follow up?

Skittyblock:
oh i thought i had switched the app to using the package instead of the cli. i'm on `0.63.3` but it seems to still flag it(?)

i believe the books app does a webview per page, but i don't know the actual performance cost of this. i assume since they do it it shouldn't be that bad, but idk

maerlin:
Pushed. Done since the last comment:

- page turns follow the finger, a pan on the web view's offset with a velocity-projected settle
- side drags off in scroll style, and a selection long press no longer turns a page through the tap zones
- scroll style reaches every edge and pads inside the document with readium's scroll padding, so content passes under the bars and there's no seam where the glass starts refracting
- taps scroll two thirds of a viewport, matching the webtoon and text readers, instead of a whole one
- chapter changes animate along the reading axis rather than always sideways
- edge effects match the other readers

The scroll spine transition is the one left. Doing it properly means a web view per document in an outer scroll view with windowing and a persisted layout cache, which is a second architecture alongside the paged one and re-opens the position restore. I would rather do it as a follow-up than grow this PR further. Your call if you want it now, let me know.

Skittyblock:
the pan gesture is better but still feels unnatural since it doesn't match with the paged image reader. i think the velocity threshold required to initiate the swipe to the next page is too high, as many swipes just scroll back to the current page. also when it activates it slows down when it should only accelerate from the current velocity. i still think using the page controller would be best to not have to try to replicate the exact behavior.

regarding the scroll transitions, that's fine. we can just leave it as is for the initial implementation.

some other things i noticed from testing:
- selecting text across multiple pages shouldn't be possible, and creates some buggy animations (would be solved by separate web views)
  - also the swiping to select normally conflicts with the page turning gesture
- the reader settings should hide the same stuff that the text reader does
- double tapping in the scroll reader seems to scroll to that position, which i think should be disabled
- the back button that shows up after selecting from the toc should probably match e.g. the auto play button with a glass style and highlight

maerlin:
> * the reader settings should hide the same stuff that the text reader does
> * double tapping in the scroll reader seems to scroll to that position, which i think should be disabled
> * the back button that shows up after selecting from the toc should probably match e.g. the auto play button with a glass style and highlight

These are fixed.

> the pan gesture is better but still feels unnatural since it doesn't match with the paged image reader. i think the velocity threshold required to initiate the swipe to the next page is too high, as many swipes just scroll back to the current page. also when it activates it slows down when it should only accelerate from the current velocity. i still think using the page controller would be best to not have to try to replicate the exact behavior.
> 
> regarding the scroll transitions, that's fine. we can just leave it as is for the initial implementation.
> 
> some other things i noticed from testing:
> 
> * selecting text across multiple pages shouldn't be possible, and creates some buggy animations (would be solved by separate web views)
>   
>   * also the swiping to select normally conflicts with the page turning gesture

For these I ran out of ideas. I tried. Can't make it work.

I tested `UIPageViewController` with one web view per page, but it fails over selectable text. Its pan gesture never begins on text pages and a pan target never fires, though swipes still work on image pages or across spine documents.

I think the cause is `WKDeferringGestureRecognizer`, which defers outside gestures for the web process but never releases over text. I verified that setting `touch-action` and `pan-y` has no effect. Disabling the recognizers only works intermittently because `WebKit` automatically re-enables them when rebuilding the content view.

Let me know if you would like us to try to fix this by working on matching the behavior by replicating. Because I can't figure out the one webview per page. I am not smart enough to figure this out. Hit a roadblock. Sorry.

Skittyblock:
text selection doesn't work at all when you have an individual web view per page in a page controller? if everything else works feel free to just push what you have and i could take a look at it

maerlin:
Page turns:
- Projection now UIScrollView's own rate. Was 5x too conservative, hence the snap-backs.
- Clamped to one page per drag.
- Settle is a native `contentOffset` animation seeded with release velocity, not a JS scrollTo from rest.
- Lift tap suppressed across the whole drag, so no double turn.

Cross-page selection: not fixed. Clamping the range on `selectionchange` works on programmatic ranges but the highlight still crosses under a finger. Reverted. Yours.

Correction: selection wasn't what broke the page controller. Selection worked. The pan never begins over text. `WKDeferringGestureRecognizer` holds it and never releases, `touch-action` does nothing. Never committed, so nothing to push.

Can't figure out how to fix the highlight to next page bug.

Skittyblock:
still think we'd be better off with a page controller than replicating the swipe gesture since:
- it still feels slightly wrong; haven't compared the two but it feels like it's too quick/rigid
- there's a flashing(?) when swiping back a page
- any swipe is activating the gesture, including swipe to dismiss and the text selection
- the ends of chapters are a little odd where you can swipe but it just insta scrolls back without an animation. we probably want to be able to transition between chapters like the text reader, but this could maybe come later since we likely need to handle loading better

from more testing:
- it probably shouldn't just show the first page while it loads everything, since you can swipe here and it messes up your current position, and is also not indicated anywhere that pages are still loading. user interaction can be disabled while a loading indicator is shown until every page is loaded (and it shouldn't incrementally update the page count, it should just set it once after fully loaded)
- you can select some text and it'll show the menu, then dismiss and the menu will stay stuck in the position for a second before disappearing abruptly (not sure if this is fixable)
- downloading epubs from komga seems to fail completely (haven't looked into this at all)
- adjusting font size/style/etc. and switching readers doesn't keep your position super well i don't think; ideally we would want the text currently at the top of the page to remain that the top of the page (i think this is what other readers do), but it seems to shift up or down. i imagine this is probably tough to handle though
- it might be nice to be able to open links in safari views when tapping on them instead of them just turning red(?) -- also they should probably be pink to match the app theme color

maerlin:
I deep dived into the `webview` mystery. Read a few docs.

#### Premise:
1. **Reflowable** EPUBs, being a collection of single XHTML documents (usually per chapter) are formatted via CSS Multi-Column. Since HTML elements flow continuously across column boundaries, an HTML document cannot be split into separate live webviews per visual page (efficiently, it's not impossible tho).
2. When embedding `WKWebView` inside `UIPageViewController`, `WKDeferringGestureRecognizer` defers and intercepts touch events over selectable text. This prevents `UIPageViewController`'s transitioning from `.possible` to `.began`, causing swipe page-turns to fail over text.

#### Readium & Apple Books approach:
• Readium (`readium/swift-toolkit`): Uses one `WKWebView` per spine item and navigates pages via CSS column offsets (more or less what we do now, needs tuning).
• Apple Books: Also loads one chapter per `WebKit` instance, but during an interactive swipe, it uses snapshot views (`snapshotView` / `layer` snapshots) to animate the turn smoothly (`slide` or `curl`) without `WebKit`'s internal gesture recognizers fighting the drag. There is no CSS manipulation (that I know of). Unfortunately, Apple does this by calling `WebKit SPI`s, and Aidoku can't do that.

If the _Apple Feel_ is what you want Aidoku to have, then we can most definitely try to replicate the Apple Books _way_ using `UIKit` APIs. This won't bring `one-webview-per-page` to life, but should be consistent with what Apple Books does.

Once this is sorted, then I can focus on fixing the other specific issues you mentioned.
But I need your call here. Because I can also simply keep tuning the current implementation.

Skittyblock:
> This won't bring `one-webview-per-page` to life

?

i did a quick look for examples of open source apps using uipageviewcontroller for epubs and found a few things, one being [kmreader](https://github.com/everpcpc/KMReader/blob/main/KMReader/Features/Reader/Views/Epub/WebPubPagedCurlView.swift). code isn't very pleasant to read but i downloaded the app and it seems to work as expected; uses a uipageviewcontroller and doesn't seem to have any issues with layouting epubs. it does have issues with the selecting text activating page swipes, which is maybe what you were saying with the snapshots(?), but i think that's tolerable since everything else is fine.

i think doing something like this would be best. you could use that app as a reference, or i'm sure there are [many others](https://github.com/search?q=epub+UIPageViewController+language%3Aswift&type=code).

maerlin:
I use `.scroll`, whose pan never begins over selectable text. KMReader uses `.pageCurl`. Using that animation costs the selection, which you already saw, and memory.
Each child loads the whole chapter into its own web view and scrolls to one sub page, so every retained page holds a full copy of the chapter.
I am testing one web view per spine document and then page it with CSS columns: https://github.com/maerlin/Aidoku/tree/experimental/epub-reader-uikit

I think if the direction of Aidoku is doing one webview per page, I disagree and can't help that much anymore.

Skittyblock:
> I use `.scroll`

not quite sure what you're referring to since you're not using a page controller(?), but kmreader has a selectable page animation, which is what the end goal would be in aidoku (e.g. to have page turn animations like the books app).

> and memory

what's the actual difference? as far as i'm aware, uipageviewcontroller is pretty good at handling the page allocation nicely. from my understanding, the expensive part is the initial layout and calculation related to that, which could happen in a single (maybe off-screen) web view, then all the other web views have to do is display the page for a given offset.

if this actually has a noticeable impact on performance i'd be willing to reconsider, but i find it somewhat hard to believe that the apple books app, which seems to work perfectly from my perspective, would choose this route if that were the case.

maerlin:
I think I can't make my point come through.
But basically either `.scroll` (via `UIPageViewController`, or our current implementation) or .pageCurl both hold the entire document in memory when showing one page. That's just the way .epubs are: every .epub is a collection of .xhtml documents. One spine is one .xhtml, then gets divided in columns (pages) when flowing, given constraints (font size, view, etc etc). There is no concept of one page that's easily callable, it's just the entire document and the column we're currently viewing.
`.pageCurl` since it needs a 3D animation, basically makes a texture of the current view (let's call it a snapshot) so it becomes its own thing, but behind it every page turn (included the preloaded ones) still hold the entire spine in memory.
`.scroll` doesn't get the same call, because it doesn't use the 3D animation. I _think_ Apple Books uses snapshots (closed source, can't tell you for real but the crumbs are there).
The fact other readers use `UIPageViewController` wasn't my question or my (negative) statement. It's the fact that if you want to use it in Aidoku, the animations needs to go through snapshots (or eventually the 3D animation).
It is doable currently for `.scroll` using pools and snapshots. The point is it will still not be one webview per page because behind the snapshot every page will still hold a webview with the entire spine in it, not just one page.

Skittyblock:
i'm not quite sure where your snapshots stuff comes in... isn't that if you're handling the animations yourself? UIPageViewController handles that all for you, regardless of the page animation you have set. or are you just explaining that's what happens in the background?

but yeah, each web view would contain the current spine document. there should only be ~3 web views at a time though pretty much. are you saying that's a problem?

maerlin:
You have to take a snapshot and make that move with the proper animation (page controller), otherwise WKWebView takes over and tries to move to the next column of the document.
At least that's what I am understanding how it works. First tries I made at disabling WKWebView scroll and using UIPageViewController, it disabled the animations in general. So probably (I assume) something like: disable wk scroll - snapshot - animate the snapshot - pull the other page from the pool - show it.

maerlin:
I am working on a separate branch to overhaul this.
Honestly at this point I am not sure this PR will ever be on the same page (pun intended) with the other readers.

Skittyblock:
i reverted the initial epub code since i may end up pushing an update before this pr lands, and i would rather wait for the implementation of epub reader here. you can probably revert that commit in this branch to fix conflicts

maerlin:
Rebuilt on UIPageViewController, one page per view controller, each page's web view holding its spine document in columns with touches off. The custom pan is gone.

- swipes and taps turn pages like the paged text and image readers, no drift, no double turn
- the book is measured behind a cover before anything shows, the total set once
- text selection: long press selects, native handles and callout; a selection cannot cross a page or turn one
- links open in a Safari view and use the app tint
- rotation and font changes rebuild behind a frosted snapshot and restore the anchor
- komga epub downloads fixed, and downloads of multi-document books now finish
- iPad: web views forced to mobile content mode, desktop mode counted whole documents as one page
- merged main and reverted the epub removal on this branch, as you suggested

Skittyblock:
ok thank you this is feeling much better now. 

in terms of behavior changes:
- it seems like keyboard arrow keys don't work to change pages
- the progress view when loading is kinda big. haven't looked at the code for this, but could we reuse the existing loading indicator? otherwise probably make it the same size
- the expanded image view is kinda odd
  - transitioning makes the whole reader view fade so i can see the series view behind it for a brief second(?)
  - the x mark button should probably be just like the navbar button so it feels more native. haven't checked the code, but we should either use a regular navbar or just use the glass effect from the back and auto play buttons

i will try and take a closer look at the code soon. after a brief glance:
- all the new epub tests could be stuck in a `Epub` folder
- stuff like `EpubChapterCache` shouldn't throw `SourceError`s; you could define a simple error enum in the object namespace, then in the calling class map it to a source error. you could even just implement localized description (i think there's a trait for this?) on the error to have it return the localized strings you're using to make it simple

Skittyblock:
another bug after using it a bit more: if you keep swiping pages forward continuously, the pages start coming in blank and not loading until you stop (this could be a debug build only thing)

maerlin:
I'll just wait till you take a longer look to start working on stuff.
This was a sort of "first pass" to be sure it was the direction you wanted.

> another bug after using it a bit more: if you keep swiping pages forward continuously, the pages start coming in blank and not loading until you stop (this could be a debug build only thing)

For this, I need to know if it's iPad or iPhone (or both). iPad, with the double columns, needs a bit more care to allow the pool to pickup pages correctly (and manage memory).

Skittyblock:
i've only checked on an iphone sim on 26.4

maerlin:
> i've only checked on an iphone sim on 26.4

Ok cool. When you feel like you've looked at it enough, let me know. I'll iron out the bugs you find.

Aidoku/Core/Sources/BuiltIn/Kavita/KavitaSource.swift
```diff
@@ -230,6 +235,29 @@ actor KavitaSourceRunner: Runner {
...
+        }
+        func makeRequest() throws(SourceError) -> URLRequest {
+            var request = URLRequest(url: url)
+            guard helper.authorize(request: &request) else {
+                throw SourceError.message("NOT_LOGGED_IN")
+            }
+            return request
+        }
+        do {
+            return try await EpubChapterCache.pages(request: try makeRequest(), sourceKey: sourceKey, chapterId: "\(chapterId)")
```
Skittyblock, line 252: i think that instead of expecting sources to return a parsed list of pages using zipUrl, we should add a new page content type like `file` that accepts cbz/zip/epub network urls, then also have a trait that allows providing the request (e.g. for adding authentication). then the app should handle parsing things rather than doing it inside kavita and komga. this way external sources will also be able to provide files in the same way

however, this is probably a refactor i should do myself separately since you're just doing the same thing as zip files, which is okay for now

Aidoku/Core/Downloads/DownloadManager.swift
```diff
@@ -66,6 +66,8 @@ actor DownloadManager {
         let archiveURL = directory.appendingPathExtension("cbz")
         if archiveURL.exists {
             return LocalFileManager.shared.readPages(from: archiveURL)
+        } else if let epubFile = directory.contents.first(where: { $0.pathExtension.lowercased() == "epub" }) {
```
Skittyblock: should we not skip the chapter directory with epub files and have them inside the series directory directly, just like the cbz files? i don't think there's any case we'd want more files in the chapter directory if we have one epub, so it doesn't make sense to have it

Aidoku/Core/Downloads/DownloadTask.swift
```diff
@@ -510,7 +564,7 @@ extension DownloadTask {
 
                 try FileManager.default.moveItem(at: tmpDirectory, to: directory)
 
-                if AppSettings.downloads.compress.get() {
+                if AppSettings.downloads.compress.get() && pages.first?.isEpubPage != true {
```
Skittyblock: don't you already check if the first page is an epub and return above? don't think this check is necessary

Aidoku/Core/Epub/EpubChapterCache.swift
```diff
@@ -0,0 +1,163 @@
...
+// downloads a server-side epub once so it reads through the same pipeline as an imported one
+enum EpubChapterCache {
+    static func fetch(request: URLRequest, sourceKey: String, chapterId: String) async throws -> URL {
+        guard let root = directory else {
+            throw SourceError.message("EPUB_DOWNLOAD_FAILED")
+        }
+        let file = root
+            .appendingPathComponent(sanitized(sourceKey), isDirectory: true)
+            .appendingPathComponent("\(sanitized(chapterId)).epub")
+        if FileManager.default.fileExists(atPath: file.path), await !isStale(request: request, file: file) {
```
Skittyblock: we have helper functions
```suggestion
        if file.exists, await !isStale(request: request, file: file) {
```

Aidoku/Core/Epub/EpubChapterCache.swift
```diff
@@ -0,0 +1,163 @@
...
+            return file
+        }
+
+        // a partial file would be served as the book, and no retry would get past the cache
+        do {
+            let response = try await URLSession.shared.download(for: request, to: file)
+            if let lastModified = (response as? HTTPURLResponse)?.value(forHTTPHeaderField: "Last-Modified") {
+                try? lastModified.write(to: lastModifiedSidecar(for: file), atomically: true, encoding: .utf8)
+            } else {
+                try? FileManager.default.removeItem(at: lastModifiedSidecar(for: file))
```
Skittyblock:
```suggestion
                lastModifiedSidecar(for: file).removeItem()
```

Aidoku/Core/Epub/EpubChapterCache.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+        markAccessed(file)
+        evict(in: root, capacity: capacity, keeping: file)
+        return file
+    }
+
+    static var directory: URL? {
+        FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first?
+            .appendingPathComponent("EpubCache", isDirectory: true)
+    }
```
Skittyblock: also, `directory`, `totalSize`, `capacity`, `CachedBook`, etc. should be at the top above all the functions
```suggestion
    static let directory: URL = FileManager.default.cachesDirectory
        .appendingPathComponent("EpubCache", isDirectory: true)
```

Aidoku/Core/Epub/EpubPageIndex.swift
```diff
@@ -0,0 +1,97 @@
+//
+//  EpubPageIndex.swift
+//  Aidoku
+//
+//  Created by Pietro Baiguini on 8/13/26.
```
Skittyblock, line 5: could put this in `EpubChapterCache` as well to keep headers consistent, but i'll probably unify all file headers at some point in the future so it doesn't really matter

Aidoku/Core/Epub/EpubPageIndex.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    private var counts: [Int?]
+
+    // both zero-based
+    struct Position: Equatable {
+        let document: Int
+        let page: Int
+    }
+
+    init(spinePaths: [String]) {
```
Skittyblock: init should be after the computed vars

Aidoku/Core/Epub/EpubPageIndex.swift
```diff
@@ -0,0 +1,97 @@
...
+//  EpubPageIndex.swift
+//  Aidoku
+//
+//  Created by Pietro Baiguini on 8/13/26.
+//
+
+import Foundation
+
+// prefix sums over counts that arrive one document at a time. every answer is nil rather than
+// approximated while they are incomplete: a total that grows is honest, a position that moves is not
```
Skittyblock: i don't really understand this. "a total that grows is honest" meaning what, that the total only increases while spine page counts are determined? i don't really see language like this

Aidoku/Core/Epub/EpubPaginationSettings.swift
```diff
@@ -0,0 +1,255 @@
...
+
+// readium-css fixes the injection order and it is load-bearing: before.css, the publication's own
+// stylesheets or default.css, then after.css. every injected value is viewport-independent;
+// --RS__viewportWidth stays at 100%, a pixel value surviving a rotation the 100vw columns do not
+struct EpubPaginationSettings: Equatable {
+    var columnCount: Int = 1
+
+    // UIDevice.current.orientation is .unknown until the device moves, so it cannot answer here
+    static func columnCount(for viewport: CGSize) -> Int {
+        UIDevice.current.userInterfaceIdiom == .pad && viewport.width > viewport.height ? 2 : 1
```
Skittyblock: is this something that should necessarily only be gated to ipads? generally i feel like phones in landscape can often be treated the same (considering plus size phones, and also the iphone fold that's coming which who knows what interface idiom it would count as)

maerlin:
Done:
- blank pages while swiping: a scroll sent to a web view that had just left the window sometimes waited a second and the page came up at its old offset. Page web views now stay in the window between pages. 0 blank turns in 111 after, 3 stalls in 298 before
- loading indicator matches the host's
- image preview presents over the reader, close button is the glass capsule the auto scroll button uses
- epub tests in `AidokuTests/Epub`
- `EpubChapterCache` throws its own `DownloadError`, `LocalizedError` with the same strings; Komga and Kavita catch its cases

arrow keys: tested on iOS 26.4 and 26.5 simulators with logging, every press reached the reader and turned a page, before and after a selection (long press). Can't reproduce. Can you share more? Which mode, which settings?

Aidoku/Core/Epub/EpubRemoteResourceProvider.swift
```diff
@@ -0,0 +1,70 @@
...
+
+        let start = Date()
+        let (data, response) = try await session.data(for: request)
+
+        if let response = response as? HTTPURLResponse, !(200..<300).contains(response.statusCode) {
+            throw EpubResourceError.requestFailed(path: path, statusCode: response.statusCode)
+        }
+
+        let elapsed = Int(Date().timeIntervalSince(start) * 1000)
+        LogManager.logger.debug("EpubRemoteResourceProvider: fetched \(path) (\(data.count) bytes) in \(elapsed)ms")
```
Skittyblock: idk if we really need debug logs like this

Aidoku/Core/Sources/BuiltIn/Kavita/KavitaSource.swift
```diff
@@ -230,6 +235,29 @@ actor KavitaSourceRunner: Runner {
...
+        } catch EpubChapterCache.DownloadError.notLoggedIn {
```
Skittyblock: i think you still need to catch errors and rethrow them as `SourceError.message`s (using the `localizedDescription`) if you want to display the error to the user.

Aidoku/Core/Sources/BuiltIn/Komga/KomgaSource.swift
```diff
@@ -194,6 +201,27 @@ actor KomgaSourceRunner: Runner {
...
+        // only a chapter this source has not listed this launch, reached by resuming straight into
+        // the reader, has to ask
+        let isEpub: Bool
+        if let known = epubChapters[chapter.id] {
+            isEpub = known
+        } else {
+            let path = "api/v1/books/\(chapter.id)"
+            // a book that cannot be described is read as images, as every chapter was before epubs
+            let book: KomgaBook? = try? await helper.request(path: path, lastWorkingMirror: &lastWorkingMirrorCopy)
+            isEpub = book.map { $0.media.mediaProfile == "EPUB" && !$0.media.epubDivinaCompatible } ?? false
```
Skittyblock: could probably stick the epub check as a computed var on `KomgaBook` since it's used multiple times

Aidoku/Extensions/WebKit/WKWebView+ContentWorld.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  WKWebView+ContentWorld.swift
+//  Aidoku
+//
+//  Created by Pietro Baiguini on 8/11/26.
+//
+//  Named for what it adds rather than for the type: Extensions/WebKit/WKWebView.swift already
+//  exists, and two source files sharing a basename collide in a target's build outputs.
+//
```
Skittyblock:
```suggestion
```

Aidoku/Features/Reader/Page/MarkdownView.swift
```diff
@@ -75,7 +75,8 @@ struct MarkdownView: View {
     }
 }
 
-// loads file urls directly from disk, since the default provider only handles network urls.
+/// Loads file urls (e.g. images extracted from epubs) directly from disk,
+/// since the default provider only handles network urls.
```
Skittyblock: dunno if the rewrite as doc comment is really necessary

Aidoku/Features/Reader/Readers/Epub/EpubPagedViewController.swift
```diff
@@ -0,0 +1,467 @@
...
+
+    private(set) var currentBookPage = 0
+
+    var onPageChanged: ((Int) -> Void)?
+
+    var onWillTurn: (() -> Void)?
+
+    private let pageViewController = UIPageViewController(
+        // scroll, not pageCurl: it is what the paged text and image readers use, and the curl can
+        // become an option later without the pages changing
```
Skittyblock:
```suggestion
```

Aidoku/Features/Reader/Readers/Epub/EpubSpineRenderer.swift
```diff
@@ -0,0 +1,775 @@
...
+        webView.scrollView.alwaysBounceVertical = !settings.paged
+        webView.scrollView.isScrollEnabled = !settings.paged
+        webView.scrollView.showsVerticalScrollIndicator = false
+        webView.scrollView.showsHorizontalScrollIndicator = false
+
+        if #available(iOS 27.0, *) {
+            // a paged view does not scroll, so an edge effect there only fades the first and last
+            // lines. the horizontal pair is hidden either way, a page drag is not a scroll
+            webView.scrollView.topEdgeEffect.style = settings.paged ? .hard : .soft
+            webView.scrollView.bottomEdgeEffect.style = settings.paged ? .hard : .soft
```
Skittyblock: any reason to have this set to hard? could just always set to soft since it's hidden when in paged mode anyways

Skittyblock (changes requested):
general comment: (for aidoku's code style) vars should all be defined at the top of a class/struct/enum, followed by nested types, then init and deinit, then functions

didn't look too close at the ui code yet, i'll probably go in and modify things myself later

maerlin:
Pushed on top of main:
- epub cache: your three suggestions, `directory` non-optional, members in the house order
- `KomgaBook.isEpub` for the two checks
- debug log, header note, `pageCurl` comment and the `MarkdownView` doc comment removed; page index comment rewritten, init after the computed vars
- edge effects always `soft`
- downloads: an epub is `series/chapter.epub` beside the cbz files, no chapter directory. `ComicInfo.xml` goes inside the epub the way a cbz carries it, otherwise the downloads list loses its titles. The epub check in the compress branch went with the directory
- vars, types, init, functions order applied to every epub file, pure moves
- columns by size class, as discussed on Discord
- `file` page type: fine, the current shape stays until then

Unrelated but noticed during the fix pass:
`findComicInfo` tests `directory.pathExtension` where it means `subdirectory`, so a cbz never yields its `ComicInfo` there. Left alone, not the scope of this PR

maerlin:
One more thing on the downloads bullet. Writing `ComicInfo.xml` into the epub is "wrong", but if we lose that file we lose the source keys plus chapter number, volume, date and series title (for the download).

Options:
- inside the epub, as pushed
- a sidecar `chapter.xml` next to `chapter.epub`, book untouched, one more file per chapter
- no record for epubs: rows show the book id, a series of books lists in id order, series title depends on `CoreData`

(OPF, the book metadata, is not a reliable source of information because it has the title but none of the keys or numbering).

Sidecar looks right to me. Your call, it's your layout.

PR #1055, opened 2026-08-28
feat: restore sources from backed-up source lists

machsix:
The new sequence of tasks during restore: 

1. Restore source lists.
2. Finds missing source IDs in the restored lists and reinstall them.
 3. Query the now-installed sources.
 4. Restore settings whose keys begin with those source IDs.

### DIFF
--- a/Aidoku/Core/Sources/SourceManager.swift
+++ b/Aidoku/Core/Sources/SourceManager.swift
@@ -370,6 +370,71 @@ extension SourceManager {
 
 // MARK: - Source Management
 extension SourceManager {
+    func missingExternalSourceKeys(in keys: Set<String>) async -> Set<String> {
+        let installedKeys: Set<String> = await CoreDataManager.shared.container.performBackgroundTask { context in
+            Set(CoreDataManager.shared.getSources(context: context).compactMap(\.id))
+        }
+        return keys.subtracting(installedKeys)
+    }
+
+    /// Installs missing sources with matching entries in the currently loaded source lists.
+    ///
+    /// If multiple lists contain the same source, the newest compatible version is used.
+    func installExternalSources(
+        keys: Set<String>,
+        progressReport: @MainActor @Sendable (_ name: String, _ key: String, _ current: Int, _ total: Int) -> Void
+    ) async -> Set<String> {
+        guard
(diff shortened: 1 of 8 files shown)


Aidoku/Core/Sources/SourceManager.swift
```diff
@@ -370,6 +370,54 @@ extension SourceManager {
...
+        guard !missingKeys.isEmpty else { return }
+
+        let appVersion = SemanticVersion(appVersionString)
+        var sourcesByKey: [String: ExternalSourceInfo] = [:]
+
+        for state in sourceListStates.values {
+            guard case let .loaded(sourceList) = state else { continue }
+            for source in sourceList.sources where missingKeys.contains(source.id) && source.fileURL != nil {
+                if let minAppVersion = source.minAppVersion,
+                   SemanticVersion(minAppVersion) > appVersion {
```
Skittyblock: character code alignment like this should be avoided when possible. it can either be a single line, or put `if` and the opening brace on their own lines and have the other statements indented and on their own lines between them

you could also maybe do `source.minAppVersion.flatMap(SemanticVersion)` but idk if that's any cleaner

Skittyblock (changes requested):
this is something i was planning to look at but keep forgetting to get around to, thanks

i think a better flow though would be to actually confirm with the user if they want to install sources (given that it triggers network requests), and could probably be skipped automatically if the user is offline (using `Reachability`)

more specifically: after doing the initial restore, before showing the missing sources alert, it should show a confirm alert asking if the user wants to check for and install sources (probably show a count for how many at least). if they confirm, it can install the sources then restore the settings for them from the backup, then show the missing sources alert for the ones that were not found

machsix:
https://github.com/user-attachments/assets/479e6893-f819-4f8a-82ae-9cbd397e6bc7

I made some changes to ask for user confirmation before looking up for missing sources. Here is a test on my ipad

Skittyblock (approved):
changes i made:
- use new missingExternalSourceKeys function for final alert
- use async stream for installExternalSources (more modern swift api preferred over progress reporting closures)
- remove message changing while installing sources (i think the progress view is good enough)
- adjust localization (not good to have manual line breaks in localization strings)

PR #1056, opened 2026-08-29
feat: allow image prefetching beyond current chapter boundary

Amqx:
Allow the paged readers to prefetch past the current chapter boundary.

The paged readers can reduce the amount of time waiting on the chapter transition screen, since image loading work can begin in the previous chapter. Note that this only looks one chapter ahead, even if the user's configured prefetch limit is higher than the total number of images loaded.

### DIFF
--- a/Aidoku/Features/Reader/Page/ReaderPageView.swift
+++ b/Aidoku/Features/Reader/Page/ReaderPageView.swift
@@ -182,6 +182,8 @@ extension ReaderPageView {
         var usePageProcessor = false
         if let source {
             // only process pages if the source supports it and the image isn't downloaded
+            // note: also skips processing raw data pages (assuming final data is already provided and doesn't need to be modified)
+            //       this should probably be changed in the future to be more correct though
             if source.features.processesPages, !url.isFileURL {
                 processors.append(PageInterceptorProcessor(source: source, pageContext: context))
                 usePageProcessor = true
--- a/Aidoku/Features/Reader/Page/ReaderPageView.swift
+++ b/Aidoku/Features/Reader/Page/ReaderPageView.swift
@@ -161,6 +161,48 @@ extension ReaderPageView {
         }
     }
 
+    /// Creates the image request used to fetch a page image.
+    static func imageRequest(url: URL, context: PageContext? = nil, sourceId: String? = nil) async -> ImageRequest {
+        let source: AidokuRunner.Source? = if let sourceId {
+            await SourceManager.shared.source(for: sourceId)
+        } else {
+            nil
+        }
+        return await imageRequest(url: url, context: context, source: source)
+    }
+
+    static func imageRequest(url: URL, context: PageContext? = nil, source: AidokuRunner.Source?) async -> ImageRequest {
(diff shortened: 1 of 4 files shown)


Aidoku/Features/Reader/Page/ReaderPageView.swift
```diff
@@ -161,6 +161,48 @@ extension ReaderPageView {
...
+        let urlRequest = if !url.isFileURL, let source {
+            await source.getModifiedImageRequest(url: url, context: context)
+        } else {
+            URLRequest(url: url)
+        }
+
+        var processors: [ImageProcessing] = []
+        var usePageProcessor = false
+        if let source {
+            // only process pages if the source supports it and the image isn't downloaded
```
Skittyblock, line 184: note from before probably shouldn't be removed

Aidoku/Features/Reader/Page/ReaderPageView.swift
```diff
@@ -161,6 +161,48 @@ extension ReaderPageView {
         }
     }
 
+    /// Creates the image request used to fetch a page image.
+    static func imageRequest(url: URL, context: PageContext? = nil, sourceId: String? = nil) async -> ImageRequest {
```
Skittyblock: trying to prefer `sourceKey` over `sourceId` now
```suggestion
    static func imageRequest(url: URL, context: PageContext? = nil, sourceKey: String? = nil) async -> ImageRequest {
```

Aidoku/Features/Reader/Readers/Paged/ReaderPagedViewController.swift
```diff
@@ -441,6 +444,49 @@ extension ReaderPagedViewController {
...
+
+            guard
+                !Task.isCancelled,
+                nextChapter == self.nextChapter,
+                viewModel.preloadedChapter == nextChapter
+            else { return }
+
+            let pages = viewModel.preloadedPages
+            guard !pages.isEmpty else { return }
+            let sourceId = viewModel.source?.key ?? viewModel.manga.sourceKey
```
Skittyblock: i think it's better to just use the value we know already exists (ik i'm inconsistent with this)
```suggestion
            let sourceId = viewModel.manga.sourceKey
```

Skittyblock (changes requested):

Aidoku/Features/Reader/Page/ReaderPageView.swift
Amqx: I changed the new stuff added by me to sourceKey. I didn't touch the old stuff though (`setPage`, `setPageImage`)

Aidoku/Features/Reader/Page/ReaderPageView.swift
Amqx: restored

Aidoku/Features/Reader/Readers/Paged/ReaderPagedViewController.swift
Amqx: changed

Skittyblock (approved):

PR #1059, opened 2026-09-01
fix: make manga cover view dismissible by swipe

Amqx:
The fullscreen manga cover view could only get closed by clicking the `x` at the top. 

This PR adds an additional swipe gesture allowing it to be dismissed by swiping down.


Skittyblock (changes requested):
why not just use zoom presentation? think it'd have the same end result and would be more native feeling than just a manual drag gesture

Amqx:
changed

Skittyblock (approved):

PR #1060, opened 2026-09-01
feat: search globally from manga title

Amqx:
Tapping a manga title on the details screen now performs a global search for that title.

### DIFF
--- a/Aidoku/App/Root/TabBarController.swift
+++ b/Aidoku/App/Root/TabBarController.swift
@@ -18,6 +18,9 @@ class TabBarController: UITabBarController {
     private var previousSelectedIndex: Int?
 
     private weak var historyNavigationController: UINavigationController?
+    private weak var searchNavigationController: UINavigationController?
+
+    private let searchController = SearchViewController()
 
     private lazy var libraryProgressView = CircularProgressView(frame: CGRect(x: 0, y: 0, width: 20, height: 20))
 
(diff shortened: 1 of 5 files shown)


Skittyblock (changes requested):
i didn't think about doing something like this but i guess it's useful to have

however, two things:
1. we lose the `textSelection` ability with this (being able to copy the title). this is kinda unfortunate but i guess you can copy the text in the search bar. so you can probably just take the property off the title text
2. when searching i think it's not setting `isLoading` properly(?) since it's displaying "no result for ..." instead of the loading indicator before results load in

i also noticed the sources and languages dropdown in the search tab disappeared but that bug was my fault so i can fix it..

Amqx:
see if you like that a little better, reverted to using text with an on tap, so the copy gesture is still available. also fixed the loading thing

Skittyblock:
i think i prefer it just being a regular button, since there should be an indication when you touch that it's a button. i think it's probably possible using uikit to have a button with an interaction effect but still allows copying text on long hold, but it's not worth trying to replicate that in uikit. i'll probably take a look at it when i get around to doing the rewrite of the series view in uikit again

also there seems to be another bug:
if i tap on a title to search it, then press the x button to go back to where i was, then press the title again, it'll show the search view with no results despite the first search having results

Amqx:
reverted to a button, squashed the bug (also happened if you global search something, exit the screen, use the `x` in text box to remove the query, then re-query the same content. that's gone now too).

Skittyblock (approved):
ty

Issue #1065, opened 2026-09-04
Severe UI hang caused by Core Data SQL


Vastrilliant:
### Summary

I encountered a UI hang in previously in Aidoku 0.8.4 (Still present in the latest version 0.9) The attached UIKit-runloop hang report shows the main thread blocked for approximately 9 seconds

This hang specifically occurs when opening an entry from the chapter updates section 

I believe this issue may be caused due to the enormous amount of data my backup file has, containing around 9000+ history entries and 20000+ tracked chapters — further corroborated by the fact that seemingly no one has reported this yet, despite being a common occurrence for me.

The main thread is performing a SwiftUI view update through View.toolbar(...), which eventually calls NSManagedObjectContext.fetch() synchronously. Core Data then attempts to access Aidoku.sqlite, but the corresponding SQL queue is occupied processing a persistent history change request.

The relevant stack is:

```
SwiftUI ViewGraph update
View.toolbar(...)
NSManagedObjectContext.fetch()
Core Data
dispatch_sync
waiting for SQLQueue
```

Meanwhile, the SQL queue remains busy for almost the entire duration:

```
NSSQLPersistentHistoryChangeRequestContext
NSPersistentHistoryResult._processResult
 persistent history change processing
SQLite fetch / sqlite3_step / pread
```
The report shows the SQL thread consuming ~9.02 seconds of CPU time while the main thread remains blocked, resulting in:

UIKit-runloop-Aidoku: timeout 9030ms

# Suspected cause

A Core Data fetch is being performed synchronously as part of SwiftUI’s toolbar/view-graph update on the main thread. When the persistent-history processing is expensive, the main thread waits for the Core Data SQL queue instead of returning to the run loop, causing the application to freeze

Moving the database work out of the main-thread view update, and/or reducing the amount of persistent-history work that must be processed synchronously, should prevent this class of hang.

The attached .zip file contains both the .ips and tailspin traces, generated by iOS hang detection 

[Logs.zip](https://github.com/user-attachments/files/31819035/Logs.zip)

### Steps to reproduce

1. Open the app 
2. Select any item from the updates section

### Device and version

iPhone 13 Pro Max, iOS27.0DB4, Aidoku v0.9 build 2

Skittyblock:
i think the fix i pushed is for what was triggering it, but #1066 is why it was hanging for so long. the next build update will include both fixes so let me know if you still have any issues after trying it

PR #1066, opened 2026-09-04
fix: reduce core data remote history processing

Amqx:
a few problems with aidoku's core data history handling:

1. it only persists its history token in memory, so every startup it replayed the entire transaction history and runs deduplication over all of it in the background
     - this is really bad because iOS just kills your app off if it uses too much memory/ cpu in the background (how i found it in the first place)
2. it does it even if the user doesn't use icloud sync (which is off by force for side loaders)

this pr fixes the following:
- persists the token to disk, so later launches resume from where they left off (falls back to 24h window if missing token)
- skips processing the history entirely if icloud sync is off, then purges the history
- purges history older than 7 days when icloud sync is on since only the sync server needs it, not reader

the biggest win would just be to stop writing history if the user has icloud disabled but i haven't looked into that enough yet

i think this also addresses #1065
probably also [this](https://discord.com/channels/927984899472883723/932316297625235477/1545328405078212728)
and likely this as [well](https://discord.com/channels/927984899472883723/932316297625235477/1545317495295180873)

### DIFF
--- a/Aidoku/Core/Database/CoreDataManager.swift
+++ b/Aidoku/Core/Database/CoreDataManager.swift
@@ -32,6 +32,12 @@ final class CoreDataManager: @unchecked Sendable {
     }()
     // only accessed from remoteHistoryQueue
     private var lastHistoryToken: NSPersistentHistoryToken?
+    private var didLoadHistoryToken = false
+
+    private static let historyTokenUrl = FileManager.default.applicationSupportDirectory
+        .appendingPathComponent("historyToken.data")
+    /// How far back a cold start looks when no token has been stored yet.
+    private static let historyColdStartWindow: TimeInterval = 24 * 60 * 60
 
     private static var shouldUseiCloud: Bool {
         AppSettings.general.icloudSync.get() && FileManager.default.ubiquityIdentityToken != nil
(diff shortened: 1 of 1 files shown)


Aidoku/Core/Database/CoreDataManager.swift
```diff
@@ -222,28 +252,78 @@ extension CoreDataManager {
...
+                )
+            }
+        }
+        return lastHistoryToken
+    }
+
+    private func clearHistoryToken() {
+        didLoadHistoryToken = true
+        lastHistoryToken = nil
+        try? FileManager.default.removeItem(at: Self.historyTokenUrl)
```
Skittyblock:
```suggestion
        Self.historyTokenUrl.removeItem()
```

Aidoku/Core/Database/CoreDataManager.swift
```diff
@@ -32,12 +32,25 @@ final class CoreDataManager: @unchecked Sendable {
...
+    private static let historyRetention: TimeInterval = 7 * 24 * 60 * 60
+    /// Purging on every remote change notification would run once per save, so throttle it.
+    private static let historyPurgeInterval: TimeInterval = 60 * 60
 
     private static var shouldUseiCloud: Bool {
         AppSettings.general.icloudSync.get() && FileManager.default.ubiquityIdentityToken != nil
     }
 
     private init() {
+        self.usesCloudKitMirroring = Self.shouldUseiCloud
```
Skittyblock: if we store this, we probably need to keep it updated. you should update it in the notification for the defaults key changing, and we should also add a listener for `NSNotification.Name.NSUbiquityIdentityDidChange` probably. then we can replace the usage of `shouldUseiCloud` with it

Skittyblock (changes requested):
thanks, i've been meaning to revisit icloud sync stuff to debug the issues. this should hopefully improve it

Skittyblock (approved):

PR #1070, opened 2026-09-10
feat: refresh library, manga, and history after iCloud sync changes

zHElEARN:
## Summary

Changes coming from iCloud were imported into the database, but nothing told the interface about them, so the library, the manga view and the history list could keep showing stale data until they were reopened or refreshed manually. For example, adding a manga or reading a chapter on one device would not show up on another device that was already open.

The app now reacts to incoming sync changes as they land.

## Behavior

- The library, the manga view and the history list refresh on their own when a sync brings changes.
- The read button is recomputed, so a stale continue-reading target is no longer shown when there is nothing left to read.

## Verification

Verified with my own developer account and iCloud container on:

- MacBook Air M4, macOS 26.6.2 — Designed for iPad build
- iPhone 17, iOS 26.6.1 — debug build

- reading a chapter on one device updates the history page open on the other
- changing a manga's bookmark state on one device updates the library page on the other
- with the same manga open on both devices, changing the bookmark state on one side updates the other
- with the same manga open on both devices, marking a chapter as read on one side updates the other

## Test

- [x] builds and runs as a Designed for iPad build on my Mac
- [x] builds and runs on my iPhone
- [x] SwiftLint passes

## Demo

https://github.com/user-attachments/assets/c6836d81-3d44-4350-a56c-2da1e789a25e

https://github.com/user-attachments/assets/eff2b935-ff33-4cb7-aa01-d4dfb143b34e

https://github.com/user-attachments/assets/cd88ee73-3d0c-4360-9a4f-d60f462c3bb2

### DIFF
--- a/Aidoku/Features/Manga/MangaView+ViewModel.swift
+++ b/Aidoku/Features/Manga/MangaView+ViewModel.swift
@@ -86,9 +86,6 @@ extension MangaView.ViewModel {
                     guard let self else { return }
                     await self.loadBookmarked()
                     await self.checkForCategories()
-                    await self.loadHistory()
-                    self.chapters = self.filteredChapters()
-                    self.updateReadButton()
                 }
             }
             .store(in: &cancellables)
@@ -180,6 +177,7 @@ extension MangaView.ViewModel {
                 guard let self else { return }
                 Task { @MainActor in
                     await self.loadHistory()
+                    self.chapters = self.filteredChapters()
                     self.updateReadButton()
                 }
             }
--- a/Aidoku/Features/Manga/MangaView+ViewModel.swift
+++ b/Aidoku/Features/Manga/MangaView+ViewModel.swift
@@ -80,6 +80,19 @@ extension MangaView.ViewModel {
     }
 
     private func registerLibraryNotifications() {
+        NotificationCenter.default.publisher(for: .updateLibrary)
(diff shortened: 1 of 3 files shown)


Aidoku/Features/Manga/MangaView+ViewModel.swift
```diff
@@ -80,6 +80,19 @@ extension MangaView.ViewModel {
...
     private func registerLibraryNotifications() {
+        NotificationCenter.default.publisher(for: .updateLibrary)
+            .sink { [weak self] _ in
+                Task { @MainActor in
+                    guard let self else { return }
+                    await self.loadBookmarked()
+                    await self.checkForCategories()
+                    await self.loadHistory()
+                    self.chapters = self.filteredChapters()
+                    self.updateReadButton()
```
Skittyblock: does it need to update the history and chapters?

zHElEARN: Initially, I was focused on updating the manga page UI and hadn’t considered the history page or the `.updateHistory` notification yet, so I put all the UI updates in the `.updateLibrary` handler.

Later, when I worked on refreshing the history page, I added the `.updateHistory` notification, so you’re right that the updates here became redundant. I’ve updated the code to remove the duplication.

PR #1071, opened 2026-09-10
fix: update pillarbox layout for initial landscape orientation

Ari-03:
- Track viewport orientation during layout
- Add coverage for portrait and landscape pillarbox behavior


Skittyblock (approved):
i removed the test since i didn't think it was super necessary to have (testing ui in that way is more of a pain to maintain then it helps prevent issues), and simplified the fix; running stuff when layouting subviews should be avoided since it's called so frequently. thank you for reporting the issue

PR #1074, opened 2026-09-11
feature: add apple translate for books

DedPiht0-MGN:
Adds Apple Translate support for book chapters on iOS 18 and later. (#1075)

Readers can translate a chapter manually, enable automatic translation, and optionally translate one chapter ahead. Source and target languages are configurable. The 10 most recently used chapter translations stay in memory and are cleared when the app restarts.

Translation preserves the selected paged or continuous scrolling mode. An optional split view displays the original text above the translation with synchronized scrolling. While translation is running, a progress indicator appears over the original text.

This PR also fixes scroll jumps around chapter transitions and prevents already loaded chapters from being loaded again when scrolling back.

**Screenshots**

Translation settings:

IMG_7696 IMG_7694

Split view with the original text and translation:

IMG_7695

**Validation**

- 19 tests passed, covering caching, translation state, pagination, scrolling, and chapter navigation in split view.
- iOS build succeeded.
- Tested on a physical device running iOS 18.7.2.


Skittyblock (commented):
i think this is out of scope for aidoku. at least, the whole split view and automatic translation. what the text reader *should* support is text selection, which should allow translation of sentences and stuff. since this is supported in the epub reader (#1052), the text reader will likely shift to using the epub reader at some point. another thing i want to support in the future is dictionaries, like how they're supported for images currently. having entire chapter machine translation isn't my goal, and would rather provide tools for people to learn new languages

if you want to keep these changes in a fork yourself you are free to

> This PR also fixes scroll jumps around chapter transitions and prevents already loaded chapters from being loaded again when scrolling back.

if you want to create a separate pr with any text reader fixes that are separate from this translate feature please feel free

PR #1079, opened 2026-09-14
Use 1024px app icon in AltStore source

izaankml:
The app `iconURL` in `apps.json` points at `apple-touch-icon.png`, which is 180x180. AltStore renders app icons in the My Apps list and Browse tab at well over that on 3x displays, so the icon is upscaled and looks soft. The small round source badge uses the same image but looks fine because it is drawn much smaller.

The site already serves the same artwork as a plain 1024x1024 square at `https://aidoku.app/images/icon.png`, which is the size Delta and the official AltStore source use for their icons. This switches `iconURL` to that file.

No source-level `iconURL` is added; like the official AltStore source, the source badge is derived from the featured app's icon.

### Before

My Apps, where the upscaled icon is visibly softer than Delta's next to it, while the small source badge is sharp:

Browse tab, same icon at a slightly smaller size:


izaankml:
Closing in favor of #1080. This targeted the `altstore` branch, which is regenerated by the Update AltStore Source workflow from the template on `main`, so the change needs to go there instead.

PR #1081, opened 2026-09-14
Restore flat legacy icon PNGs in the app bundle

izaankml:
AltStore's My Apps tab shows a soft, glass-rendered version of the Aidoku icon, while Delta and AltStore itself show their flat icons there. The cause is in how Xcode compiles the icon.

### What happens

AltSign loads the installed app's icon by taking the last name in `CFBundleIconFiles` and calling `UIImage(named:in:)` on the bundle, which resolves to the loose `AppIcon60x60@2x.png` file. Since 0.9 is built with an Icon Composer `AppIcon.icon`, Xcode renders that file's glass appearance into the loose PNG, at 120px only. Delta 1.7.1 and AltStore 2.2.2 were built with Xcode 16, so their loose PNGs are plain.

The project already sets `--enable-icon-stack-fallback-generation=disabled` in `ASSETCATALOG_OTHER_FLAGS`, but it has no effect. When a `.icon` file and an `.appiconset` share a name, actool drops the icon set entirely: the compiled catalog contains none of its sized entries, only the 1024px images rendered from the icon stack. With no legacy items left, actool forces fallback generation back on. I verified this with Xcode 26.3 by running actool directly with the flag in every spelling, and by compiling the catalog with and without the `.icon` file.

### Change

A "Restore Legacy App Icons" build phase, placed after Copy Bundle Resources, copies the flat 120, 180, and 152 px icons from `AppIcon.appiconset` over the generated `AppIcon60x60@2x.png`, `AppIcon60x60@3x.png`, and `AppIcon76x76@2x~ipad.png`. The 3x file was not previously emitted, so 3x devices no longer upscale the 120px image.

The iOS 26 icon stack and the flattened 1024px icons inside `Assets.car` are untouched, so the home screen, Spotlight, and Settings icons do not change on any iOS version. Only readers of the loose PNGs see the difference: AltStore, SideStore, and similar app managers.

Verified with a Release build for generic iOS: the phase runs, the three bundle PNGs are byte-identical to the asset catalog sources, and `Assets.car` still contains the icon stack layers.

### How it looks in AltStore today

Related: #1080 fixes the separate source-level `iconURL` used by the Browse tab.


Skittyblock:
i suppose this can be done. i assume this also fixes the app icon for older ios versions, which is something that bothered me when xcode removed `enable-icon-stack-fallback-generation`. i will have to test this, but we can probably remove that flag now (as it's useless anyways)

PR #1082, opened 2026-09-14
chore: correct altstore markdown formatting

izaankml:
The What's New section for Aidoku in AltStore shows stray bullets inside words and a raw markdown link. The cause is the markdown cleanup in the AltStore source script.

### What happens

`update_altstore_json.py` builds each version's `localizedDescription` from the GitHub release body. Its cleanup step replaced every hyphen in the text with a bullet, not only the list markers at the start of each line, so "built-in" rendered as "built•in" and "Yomitan-compatible" as "Yomitan•compatible". The same substitution broke the release notes URL, and markdown links were left as `[label](url)`.

### Change

The cleanup is replaced with a markdown-to-plain-text conversion that:

- treats `-`, `*`, or `+` as a bullet only at the start of a line
- rewrites `[label](url)` and `[label](<url>)` as `label (url)`
- unwraps `<https://...>` autolinks and strips other HTML tags
- strips bold, italic, strikethrough, and heading markers
- normalises line endings and collapses runs of blank lines

Verified against the v0.9, v0.8.4, and v0.8.3 release bodies. The output for v0.9 was published as a temporary test source and checked in AltStore 2.2.2.

Note that AltStore does not enable link detection on the version description text view, so the URL is shown as text but is not tappable. That is an AltStore limitation and applies to every source.

### Before and after

 

The already published 0.9 entry on the `altstore` branch still has the old text. Running the Update AltStore Source workflow manually after merging regenerates it.


Skittyblock (approved):
thanks

PR #1091, opened 2026-09-23
fix: prevent search crash on macOS

demirozg:
## Summary

Fixes a reproducible crash when opening the Search tab on macOS.

The crash occurs inside UIKit while presenting `UISearchController`, with the stack reaching `_screenBasedFocusUnsupported`.

## Fix

Add a macOS-specific `UIApplicationSceneManifest-macos` configuration with:

```xml
<key>UIApplicationSupportsMultipleScenes</key>
<true/>
```

The existing iOS/iPadOS scene configuration remains unchanged, so multiple scenes stay disabled there.

## Testing

Tested locally using the **Mac (Designed for iPad)** destination.

Before this change:

* Opening Search consistently crashed the app on macOS.

After this change:

* Search opens normally.
* The `_screenBasedFocusUnsupported` crash no longer occurs.

Also verified that the iOS/iPadOS scene manifest remains unchanged.

Cheers


demirozg:
@copilot Fix the code for all comments in [this review thread](https://github.com/Aidoku/Aidoku/pull/1091#pullrequestreview-5296693459).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

